Issue #75, opened 2020-02-27
Failure to handle some modules on Windows due to encoding


adambenali:
Here is how to reproduce the error:

0. Use Windows
1. Install `grimp`.
2. Create a python package `test_encoding`. Inside `__init__.py` add this content : `print('變量')`.
3. Open a python shell and do:
```python
>>> import grimp
>>> graph = grimp.build_graph('test_encoding')
```
Fails with : 
```
  File "C:\Users\adam\Anaconda3\lib\site-packages\grimp\adaptors\filesystem.py", line 26, in read
    return file.read()
  File "C:\Users\adam\Anaconda3\lib\encodings\cp1252.py", line 23, in decode
    return codecs.charmap_decode(input,self.errors,decoding_table)[0]
UnicodeDecodeError: 'charmap' codec can't decode byte 0x8f in position 12: character maps to <undefined>
```

seddonym:
Thanks for the bug report!

I'll look into it, though I don't have access to a Windows machine so I can't promise how easy it will be to fix. I'd welcome a pull request if you felt so inclined.

seddonym:
Actually thinking about it, would you mind uploading the file you created here? I'll see if I can reproduce the problem with that file.

adambenali:
Here you are : [test_encoding.zip](https://github.com/seddonym/grimp/files/4266766/test_encoding.zip).

seddonym:
Thanks for that. Interestingly, that file loads fine for me. My guess is that what is happening is that your file is incorrectly encoded, and it's somehow being fixed in the process of transferring to me.

I think probably the right approach is for Grimp to handle file decoding problems in a more forgiving manner, but also you might want to look into why this is happening on your machine.

Incidentally, this is a good article on the subject. http://python-notes.curiousefficiency.org/en/latest/python3/text_file_processing.html

adambenali:
Oh, that is curious. Maybe it depends on the region. What is the Windows version you are using ?
Also, could you please check the default encoding on the machine ? You can do that by typing `[System.Text.Encoding]::Default` in a PowerShell.

seddonym:
I'm not using Windows. But my system's preferred encoding is UTF-8. What do you get if you run this?
```
python -c "import locale; print(locale.getpreferredencoding())"
```

adambenali:
> I'm not using Windows.

This is precisely why it works for you.
[...]

PR #76, opened 2020-03-15
Use tokenize.open for opening files

seddonym:
Motivated by https://github.com/seddonym/grimp/issues/75.

Unfortunately this is currently causing problems with the coverage report, which is why the tests are failing.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -84,3 +84,7 @@ Changelog
 1.2 (2019-11-27)
 ----------------
 * Significantly increase the speed of building the graph.
+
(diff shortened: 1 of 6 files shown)


adambenali:
Tested it locally on Windows and it seems to fix the issue. However, looks like the version of `coverage` needs to be updated to `5.0.3`: 
https://github.com/seddonym/grimp/blob/9ab28d61df4684c360da36ee2c48cfb3e2c247c1/tox.ini#L51

seddonym:
That did the trick, thanks!

PR #78, opened 2021-01-16
Propagate ImportLibPackageFinder origin assertion

### LINKED ISSUE #53: Better handling of installable packages without __init__.py
If you have a package that's installed via a setup.py, but it doesn't have an `__init__.py` file at the top level, you get the following error:

```
...
File ".../grimp/adaptors/packagefinder.py", line 23, in determine_package_directory
    assert package_filename.origin  # For type checker.
AssertionError
```
The simple fix at the user's end is to add the missing `__init__.py` file. But it's a confusing error message.

The problem is, it's assuming that the `origin` attribute on the returned ModuleSpec will always be set. That's not true, see the docs: https://docs.python.org/3.6/library/importlib.html#importlib.machinery.ModuleSpec.origin

kamadorueda:
- image
- Fixes #53
- Fixes seddonym/import-linter#93

### DIFF
--- a/src/grimp/adaptors/packagefinder.py
+++ b/src/grimp/adaptors/packagefinder.py
@@ -21,5 +21,11 @@ class ImportLibPackageFinder(AbstractPackageFinder):
             raise ValueError(
                 "Could not find package '{}' in your Python path.".format(package_name)
             )
-        assert package_filename.origin  # For type checker.
+
+        if not package_filename.origin:  # For type checker.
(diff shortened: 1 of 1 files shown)


seddonym:
Thanks for this! I'm merging into a different branch, as I'd like to make sure there's test coverage. It also looks like checking `has_location` might be better, as namespace packages do have an origin, but it's not loadable (it's just the string `'namespace'`).

PR #81, opened 2021-05-31
Improve mypy interoperability

ZipFile:
* PEP-561
* Fix few errors when running `mypy --strict`

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,0 +1,3 @@
+[mypy]
+warn_unused_configs = True
+exclude = tests/assets/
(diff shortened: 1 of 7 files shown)


seddonym (approved):
Thanks for this, much appreciated!

Before I merge, would you like to add your name to AUTHORS.rst? (You don't have to!)

ZipFile:
I don't think this is a significant enough change to be credited.

seddonym:
> I don't think this is a significant enough change to be credited.

Fair enough - but I appreciate it anyway!

PR #87, opened 2021-12-23
Add namespace packages support (fixes #80)

### LINKED ISSUE #80: Support for namespace packages
Python provides the ability to organize a package so that it exists in multiple directory hierarchies. These are known as ['namespace packages'](https://www.python.org/dev/peps/pep-0420/#differences-between-namespace-packages-and-regular-packages).

It is not a priority to support such packages in Grimp for the time being, though pull requests would be considered!

mrcljx:
Adds support for packages that are organized in a namespace.
Does NOT allow adding a namespace to the graph (questionable
semantics, definitely a nice to have like a glob).

```
build_graph("acme.foo", "acme.bar") # yes
build_graph("acme") # no
```

Before this would fail because `find_modules` would discover `acme.foo` as `foo`.
Fixing that led to another failure, because `_determine_module_filename`
assumed that a package wouldn't contain a `.`.

- [x] Unit tests
- [x] Update docs on AbstractImportScanner
- [ ] Add Authors
- [ ] Add Changelog

Let me know if there are additional changes needed.

### DIFF
--- a/src/grimp/adaptors/importscanner.py
+++ b/src/grimp/adaptors/importscanner.py
@@ -91,9 +91,14 @@ class ImportScanner(AbstractImportScanner):
         package_directory, package_name = self._lookup_module_package_directory(module)
         package_components = package_name.split(".")
         module_components = module.name.split(".")
-        assert module_components[0:len(package_components)] == package_components, "Module should be part of package"
+        assert (
+            module_components[0:len(package_components)] == package_components
+        ), "Module should be part of package"
 
-        filename_root = self.file_system.join(package_directory, *module_components[len(package_components):])
+        filename_root = self.file_system.join(
+            package_directory,
+            *module_components[len(package_components):]
+        )
         candidate_filenames = (
             f"{filename_root}.py",
             self.file_system.join(filename_root, "__init__.py"),
@@ -101,7 +106,7 @@ class ImportScanner(AbstractImportScanner):
         for candidate_filename in candidate_filenames:
             if self.file_system.exists(candidate_filename):
                 return candidate_filename
-        raise FileNotFoundError(f"Could not find module {module} ({package_components} {module_components}).")
+        raise FileNotFoundError(f"Could not find module {module}.")
 
     def _lookup_module_package_directory(self, module: Module) -> Tuple[str, str]:
         for package_directory, (package_name, modules) in self.modules_by_package_directory.items():
--- a/src/grimp/adaptors/importscanner.py
+++ b/src/grimp/adaptors/importscanner.py
@@ -1,6 +1,6 @@
 import ast
 import logging
-from typing import List, Set
+from typing import List, Set, Tuple
 
 from grimp.application.ports.importscanner import AbstractImportScanner
 from grimp.domain.valueobjects import DirectImport, Module
@@ -88,14 +88,12 @@ class ImportScanner(AbstractImportScanner):
         Any given module can either be a straight Python file (foo.py) or else a package
         (in which case the file is an __init__.py within a directory).
         """
+        package_directory, package_name = self._lookup_module_package_directory(module)
+        package_components = package_name.split(".")
         module_components = module.name.split(".")
-        package_directory = self._lookup_module_package_directory(module)
-        package_directory_parts = self.file_system.split(package_directory)
-        assert (
-            module_components[0] == package_directory_parts[-1]
-        ), "The package directory should be the same as the first part of the module name."
+        assert module_components[0:len(package_components)] == package_components, "Module should be part of package"
 
-        filename_root = self.file_system.join(package_directory, *module_components[1:])
+        filename_root = self.file_system.join(package_directory, *module_components[len(package_components):])
         candidate_filenames = (
             f"{filename_root}.py",
             self.file_system.join(filename_root, "__init__.py"),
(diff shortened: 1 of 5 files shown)


seddonym:
Sorry for taking a while to get onto this, and thanks for taking the trouble to look into it!

I'd definitely like better support for namespace packages. I need a few days to digest this particular approach and whether it's the right direction, I'll get back to you!

seddonym:
Hi Marcel,

Thanks again for the PR. I've been considering this change a bit more. It's interesting, I had not intended that `build_graph` would be called using subpackages. It turns out it does allow it, but the behaviour is incoherent. I've written up a separate issue on that [here](https://github.com/seddonym/grimp/issues/88).

As you'll see in my comment on that issue, I am currently of the opinion that we should not allow subpackages to be passed to `build_graph`. Feel free to comment on that issue if you disagree! 

I think there might still be scope for tackling the namespaces packages issue, but I would expect that not to change the API of `build_graph` - rather it should just behave consistently with Python's import behaviour. Essentially, calling code shouldn't have to know whether or not a package is a namespace package.

For this reason, I'm probably going to close this PR - sorry.

I'm still interested to understand why you'd like namespace packages support (it has always felt like an edge case, but you're not the only one to request it). Are you using [native namespace packages](https://packaging.python.org/en/latest/guides/packaging-namespace-packages/#native-namespace-packages), and if so, why? Or do you have a dependency that is packaged in this way that is causing problems? 

Thanks in advance,

David

mrcljx:
> Sorry for taking a while to get onto this, and thanks for taking the trouble to look into it! [...] I need a few days to digest this particular approach and whether it's the right direction

Hi David, no worries at all. Hope you were able to take some time off and thanks for taking the time. I felt like this would require some vision alignment.

> Are you using native namespace packages?

Yes. One of our dependencies is `google.protobuf` where `google` is a namespace package. But we also started using organization prefixes for our package names (was: `org_foo` -> now is `org.foo`). 

> it should just behave consistently with Python's import behavior. Essentially, calling code shouldn't have to know whether or not a package is a namespace package.

I had two goals when I made the change:

1. Simplest thing possible while still being sound and maintainable.
2. Be consistent with Python's import behavior

We seem to disagree on whether (2) is true. My view is this: while importing a namespace package (e.g. `import google`) totally works(tm), it is use- and meaningless as it will only return a stub module. That's for example why I didn't touch the existing exception if one were to pass a namespace module: while legal, it's very likely a programming error, as no sub-packages will be found.

The new #88 states that only "top-level packages" should be passed. This makes sense to me (it's also how I use it) but I would suggest to tweak the definition of "top-level" to be "any module that has no parents or for which all of the parents are namespace packages" to allow for PEP420 namespace prefixes:

```
def is_top_level_module(parts):
  *prefix, last = parts
  return all(is_namespace_module(m) for m in prefix) and not is_namespace_module(last)

def is_namespace_module(module):
  return module.__spec__.origin == "namespace"
```

But I understand if you disagree here.

seddonym:
Thanks for the detailed response.

> One of our dependencies is google.protobuf where google is a namespace package. But we also started using organization prefixes for our package names (was: org_foo -> now is org.foo).

Ah, makes sense. I think Grimp does need to support this kind of thing.

> I would suggest to tweak the definition of "top-level" to be "any module that has no parents or for which all of the parents are namespace packages" to allow for PEP420 namespace prefixes.

I could definitely be persuaded. Leave it with me and I will have a think.

I will leave this PR open in the meantime as possibly we may want to build on it. Thanks again!

mrcljx:
@seddonym Just wanted to ask you whether you came to a conclusion here. I'll probably have to fork grimp and lint-imports for us internally otherwise.

mrcljx:
For reference, I'm getting this error otherwise:

```
Package tools appears to be a 'namespace package' (see PEP 420), which is not currently supported. If this is not deliberate, adding an __init__.py file should fix the problem.
```

seddonym:
Hi Marcel,

Thanks for your patience, and the nudge on this.

I think you're probably right about the direction we need to take. I'll aim to get onto this in the next week or so.

seddonym:
Thanks for your patience. I have given this a look in https://github.com/seddonym/grimp/pull/89, it is still a work in progress but it should functionally work now. I'd love to know if this meets your needs now.

seddonym:
Also... I owe you an explanation of why I opened a new PR rather than merging yours.

When I first build Grimp, I wasn't aware of namespace packages and there are assumptions in almost every corner of the codebase that I wanted to address. A lot of the contracts of the different interfaces needed to be rethought (hence why this feature has been so long in coming!) Although I haven't used your PR directly I'll be sure to credit you as a contributor. Also just the fact that you made the effort helped validate that it would be worth spending time on this.

Thanks again!

seddonym:
I'm going to close this pull request in favour of https://github.com/seddonym/grimp/pull/89.

Would appreciate any thoughts on the implementation (no need to do an in depth review of the code). In particular, does this meet your needs?

mrcljx:
> When I first built Grimp, I wasn't aware of namespace packages and there were faulty assumptions in almost every corner of the codebase that I wanted to address. A lot of the contracts of the different interfaces needed to be rethought (hence why this feature has been so long in coming!) Although I haven't used your PR directly I'll be sure to credit you as a contributor. Also just the fact that you made the effort helped validate that it would be worth spending time on this.

No worries, I felt like a more invasive change would be needed but that's not something I would want to do as an external contributor. So thanks for overthinking it and taking it over the finish line. 👍 The implementation in #89 looks good to me (although I for sure am not 100% familiar with the codebase) and after upgrading grimp to 1.3 I was able to tell `import-linter` to use some namespaced packages as roots (will report back if that somehow doesn't work).

seddonym:
Great. Helpful to know you are using it for Import Linter, I'm planning to do an Import Linter release soon with explicit support for namespace packages, so I'll tag you on that.

PR #99, opened 2022-12-22
Fix tests for windows platforms

yakMM:
The tests are failing on windows because they were written for linux with a slash file path separator while windows uses backslash.

This PR aims at making the tests run on all platforms, unix-based and windows alike.

Modification:

1. `modulefinder` has a call to `os.sep`. This is platform dependent. I believe it make sense to add the separator in the AbstractFileSystem construct, so the FakeFileSystem used for tests can return a slash ("/") separator even when running on windows (as all tests and the FakeFileSystem  were written with UNIX path in mind)
2. In `test_packagefinder.py`, `test_determine_package_directory` uses the "assets" folder, which contains real files, but FakeFileSystem is given to the function under test. I believe the function should use the real grimp FileSystem here, as the assets folder is real, not fake.

The following PR make the tests pass on windows with minimal changes, I rely on github CI to check linux platforms.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -10,7 +10,7 @@ on:
 
 jobs:
   tests:
-    name: tox on ${{ matrix.os }}, python ${{ matrix.python-version }}
+    name: tox on python ${{ matrix.python-version }}, ${{ matrix.os }}
     runs-on: ${{ matrix.os }}
 
     strategy:
--- a/src/grimp/adaptors/filesystem.py
+++ b/src/grimp/adaptors/filesystem.py
@@ -10,6 +10,10 @@ class FileSystem(AbstractFileSystem):
     Abstraction around file system calls.
     """
 
+    @property
+    def sep(self) -> str:
+        return os.sep
+
     def dirname(self, filename: str) -> str:
         return os.path.dirname(filename)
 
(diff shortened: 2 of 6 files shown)


yakMM:
Additionally, it might be interesting (or not) to add a windows CI (and a macos one?)

seddonym:
Thanks for this - yes I think it would be a good idea to add a failing test to demonstrate the problem first, if you don't mind.

yakMM:
@seddonym It's more about the current tests failing on windows platforms, see https://github.com/seddonym/grimp/pull/102 in which I added windows CI

yakMM:
I added windows CI in this pull request as well.

So (if I did everything properly), you will see https://github.com/seddonym/grimp/pull/102 CIs failing and https://github.com/seddonym/grimp/pull/99 CIs succeding

seddonym (changes requested):
Thanks for taking a look at this, a good improvement.

Do you know if there is a simple way to distinguish between the different Github actions descriptions, at the moment there are a lot of identical names. If you're not sure I can look into it at some point.

yakMM:
Yes the names are confusing atm, good point.

It is no problem to change, see last commit, I went with `tox on ${{ matrix.os }}, python ${{ matrix.python-version }}`.

Tell me if you prefer another wording

seddonym (approved):
Looks great! Thanks for doing this.

PR #102, opened 2022-12-22
DO NOT MERGE Add macos and windows to CI

yakMM:
Add windows and macos targets to github actions.

This demonstrate why https://github.com/seddonym/grimp/pull/99 is needed: I expect windows tests to fail

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -10,7 +10,7 @@ on:
 
 jobs:
   tests:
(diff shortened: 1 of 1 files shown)


yakMM:
This PR is for demonstration purposes only, https://github.com/seddonym/grimp/pull/99 should be merged instead

seddonym:
Thanks for the demo PR, closing as the intended PR is now merged.

Issue #38, opened 2018-12-20
Better support for multiline import statements


seddonym:
```
>>> g.get_import_details(importer='django.core.checks', imported='django.core.checks.messages')
[{'importer': 'django.core.checks', 'imported': 'django.core.checks.messages', 'line_number': 1, 'line_contents': 'from .messages import ('}]
```

PR #103, opened 2022-12-27
Annotate get_import_details with a TypedDict

### LINKED ISSUE #94: Annotate ImportGraph.get_import_details with a TypedDict
It would be great to improve the type annotations on [get_import_details](https://grimp.readthedocs.io/en/stable/usage.html#ImportGraph.get_import_details) by using a [TypedDict](https://peps.python.org/pep-0589/#class-based-syntax).

This will help downstream packages such as Import Linter better work with the results of that method.

Note: TypedDict is only available in `typing` from Python 3.8, so for the time being we'll need to import it from `typing_extensions`.

mwgamble:
This is a big improvement for all code that utilises the get_import_details() method on a graph object. Type-checkers can now verify which keys are being accessed, rather than just allowing any access on any key. They also natively understand the type of each supported key, without us having to cast values manually.

As this project needs to support Python 3.7 for the time being, this introduces a dependency on `typing_extensions` to utilise TypedDict.

Fixes #94

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -55,4 +55,8 @@ setup(
         'Programming Language :: Python :: Implementation :: CPython',
         'Topic :: Utilities',
     ],
+    python_requires=">=3.7",
+    install_requires=[
+        "typing-extensions>=3.10.0.0",
+    ],
 )
--- a/src/grimp/adaptors/graph.py
+++ b/src/grimp/adaptors/graph.py
@@ -1,5 +1,5 @@
 from copy import copy
-from typing import Any, Dict, List, Optional, Set, Tuple, Union, cast
+from typing import Dict, List, Optional, Set, Tuple, cast
 
 from grimp.algorithms.shortest_path import bidirectional_shortest_path
 from grimp.application.ports import graph
@@ -21,7 +21,7 @@ class ImportGraph(graph.AbstractImportGraph):
         self._edge_count = 0
 
         # Instantiate a dict that stores the details for all direct imports.
-        self._import_details: Dict[str, List[Dict[str, Any]]] = {}
+        self._import_details: Dict[str, List[graph.DetailedImport]] = {}
         self._squashed_modules: Set[str] = set()
 
     # Dunder methods
@@ -134,8 +134,8 @@ class ImportGraph(graph.AbstractImportGraph):
                 {
                     "importer": importer,
                     "imported": imported,
-                    "line_number": line_number,
-                    "line_contents": line_contents,
+                    "line_number": cast(int, line_number),
+                    "line_contents": cast(str, line_contents),
                 }
             )
 
@@ -235,7 +235,7 @@ class ImportGraph(graph.AbstractImportGraph):
 
     def get_import_details(
         self, *, importer: str, imported: str
-    ) -> List[Dict[str, Union[str, int]]]:
+    ) -> List[graph.DetailedImport]:
         import_details_for_importer = self._import_details.get(importer, [])
         # Only include the details for the imported module.
         # Note: we copy each details dictionary at this point, as our deepcopying
--- a/src/grimp/application/ports/graph.py
+++ b/src/grimp/application/ports/graph.py
@@ -1,5 +1,14 @@
 import abc
-from typing import Dict, Iterator, List, Optional, Set, Tuple, Union
+from typing import Iterator, List, Optional, Set, Tuple
+
+from typing_extensions import TypedDict
+
+
+class DetailedImport(TypedDict):
+    importer: str
+    imported: str
+    line_number: int
+    line_contents: str
 
 
 class AbstractImportGraph(abc.ABC):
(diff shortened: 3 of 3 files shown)


seddonym:
Fixes https://github.com/seddonym/grimp/issues/94.

seddonym (commented):
Thanks for this, I totally agree, it's a big improvement.

My concern with merging as-is is the use of a plural for the type, which I find leads to unidiomatic code, particularly when dealing with collections of the type (what do you call a list of ImportDetails?). To be honest this is just exposing an existing sloppiness of naming, where sometimes import details refers to one import, sometimes to multiple. I'd like to address this before exposing it as a public type.

The best option I can think of is `ImportDetail`, which I don't like that much, so I'd like to see if we can think of a better option. Any ideas?

mwgamble:
I chose `ImportDetails` because the method is named `get_import_details`. I agree though, it's not a great name.

In lieu of something better, how about `get_import_data` and `ImportData`? The plural would be `imports_data`.

mwgamble:
It only occurred to me when reading through #94 just now that `TypedDict` is only available in Python 3.8 and above. As Python 3.7 will be completely EOL in less than six months, and to avoid the unnecessary import from `typing_extensions`, now would be a good time to switch to using dataclasses. Returning a dataclass would ultimately be better (regular objects are much more efficient than dictionaries).

How do you feel about collating a few breaking changes and calling it v3? I think there's other types that could also benefit from being converted to proper classes. It would also allow removing the `typing_extensions` dependency in Import Linter.

seddonym:
I share your love of dataclasses, but I also like the simplicity of an API that just returns native datatypes.

Ultimately I think it's just a matter of personal preference, and I'm not convinced it's worth the breaking change in this case.

Let's keep this ticket focused on the non-breaking type annotation, which I don't think has any downsides. Feel free to open a separate ticket for changing the API return value to dataclasses if you want to keep the idea open.

djmattyg007:
> I share your love of dataclasses, but I also like the simplicity of an API that just returns native datatypes.
>
> Ultimately I think it's just a matter of personal preference, and I'm not convinced it's worth the breaking change in this case.

For what it's worth, it's not just a matter of what different people consider to be *simple*. Using dictionaries has a major downside: they're mutable, and therefore not hashable. This means they cannot be stored in sets, for example. On the other hand, a dataclass can trivially be made immutable (`@dataclass(frozen=True)`).

> Let's keep this ticket focused on the non-breaking type annotation, which I don't think has any downsides.

Sure thing. Could I please get a comment on the use of `TypedDict` and support for Python 3.7? It impacts where I take this PR next.

> Feel free to open a separate ticket for changing the API return value to dataclasses if you want to keep the idea open.

Sounds good :+1:

seddonym:
> Could I please get a comment on the use of TypedDict and support for Python 3.7? It impacts where I take this PR next.

I think it's fine to import it from `typing_extensions`.

Naming-wise, `DetailedImport` is my current favourite. What do you think?

djmattyg007:
> I think it's fine to import it from typing_extensions.

Just to be clear, your preference is to keep support for Python 3.7 for the time being, and add a dependency on `typing_extensions`? I'm happy to roll with that, I just want to confirm that you'd prefer to do that even though Python 3.7 is EOL soon.

> Naming-wise, DetailedImport is my current favourite. What do you think?

My only objection to using the word "detailed" is that the returned data isn't particularly detailed (as acknowledged by #38).

seddonym:
> Just to be clear, your preference is to keep support for Python 3.7 for the time being, and add a dependency on typing_extensions? I'm happy to roll with that, I just want to confirm that you'd prefer to do that even though Python 3.7 is EOL soon.

Yes please!

mwgamble:
@seddonym I've updated the PR as per your recommendations. Could you please approve the workflow run? I don't have a good way to test with Python 3.7 locally.

Also, I was wondering how long you intend to maintain support for Python 3.7? Is it just until it goes EOL later this year?

seddonym:
> Also, I was wondering how long you intend to maintain support for Python 3.7? Is it just until it goes EOL later this year?

Yes exactly.

mwgamble:
@seddonym I've pushed a fix. It appears I left in an import of `typing.Union` by accident.

seddonym:
Thanks, I've relaxed the restrictions on running actions so 🤞🏻 workflows will run automatically from now on.

mwgamble:
@seddonym Looks like the build has passed successfully :) Do you need anything more from me to get this merged?

seddonym:
Thanks so much for your help!

djmattyg007:
@seddonym how soon do you think you could cut a new release with this improvement? I'd like to implement a PR for Import Linter to make use of the new type, but it's difficult for me to do that before the change has been released here.

seddonym:
It is done. Enjoy!

PR #100, opened 2022-12-22
Conditional imports

yakMM:
Draft PR for https://github.com/seddonym/grimp/issues/64

Are left to do:

- [x] Feature implementation
- [ ] Discuss breaking changes
- [ ] Write new tests
- [x] Update documentation

PR #122, opened 2023-08-02
Add conditional imports to import graph class

jthevos:
This PR is a notional POC attempting to capture the following scenario.
    if TYPE_CHECKING:
        from package import module

    This is motivated by wanting to lint for certain forbidden imports but allow the contracts to be kept if the if condition evaluates to false.

### DIFF
--- a/src/grimp/adaptors/graph.py
+++ b/src/grimp/adaptors/graph.py
@@ -156,6 +156,51 @@ class ImportGraph(graph.ImportGraph):
         self._importeds_by_importer.setdefault(imported, set())
         self._importers_by_imported.setdefault(importer, set())
 
+    def add_conditional_import(
+        self,
+        *,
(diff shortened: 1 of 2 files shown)


seddonym:
Thanks for looking at this!

There is actually already [a draft PR](https://github.com/seddonym/grimp/pull/100) to add metadata relating to whether or not the import is within an `if TYPE_CHECKING` block, take a look! It seems to have stalled but I don't think it's that far off.

I'm not sure about expanding this to conditionals in general. I suppose adding a flag 'is_conditional` to each import _could_ be a way to go, but other than for `if TYPE_CHECKING`, I'm not sure I can think of a genuine use case. Feel free to propose one.

seddonym:
Superceded by https://github.com/seddonym/grimp/pull/133.

Issue #64, opened 2019-08-05
Add support for detecting whether an import is only made during type checking


antdking:
Specifically, be able to detect when imports are used under `typing.TYPE_CHECKING`.

Raising this in Grimp, as it appears the functionality isn't in place, and so can't be implemented with a custom contract.

While I understand that this is solved by introducing an interface, this results in a lot of boilerplate code when all you're seeking is basic static analysis + hints in an editor.

Given a Contract:
```ini
name = Prevent cross-reliance on API Clients
type = independence
modules =
    ulama.api_client_1
    ulama.api_client_2
    ulama.remote_transfer
```

and modules of:
```py
# ulama/api_client_1.py
class ApiClient1: ...

# ulama/api_client_2.py
class ApiClient2: ...

# ulama/remote_transfer.py
from typing import TYPE_CHECKING
  from .api_client_1 import ApiClient1
  from .api_client_2 import ApiClient2

class RemoteTransfer:
  from_client: ApiClient1
  to_client: ApiClient2
```

I would expect this to pass the contract.

to reiterate, understand it'd be better to introduce a stop gap, but feel there's a use-case for keeping things simple in the code, while still being able to offer an external contract to prevent mistakes.

seddonym:
Hi Anthony,

Thanks for this - and sorry for the delay in responding.

Am I right in thinking the code sample should read:

```
...
from typing import TYPE_CHECKING

if TYPE_CHECKING:
  from .api_client_1 import ApiClient1
  from .api_client_2 import ApiClient2
...
```

I think potentially it's a good idea, though I would lean towards it not being the default behaviour, as arguably if something knows about a type, then it's a dependency.

Happy to look at a PR if you're eager for this functionality to get included, if not I'll just leave it as an open issue for now.

blokje:
@seddonym This issue seems stale, any idea if you will ever import this or can you give a pointer on where to start?

We are looking to implement `import-linter` but use `if TYPE_CHECKING:` all over the place. We have (at this moment) no way to implement import-linter in a strict enough manner.

seddonym:
Hi @blokje, I'm not actively planning to work on this but I would be happy to consider a PR (or two).

Perhaps the best way to approach it would be as follows.

First - as a general principle this should not be default behaviour for Grimp nor for Import Linter.

## Grimp

Perhaps something like this:

```
>>> graph.get_import_details(
    importer='myproject.foo',
    imported='myproject.bar',
    include_type_checking=True,
)

[
[...]

PR #130, opened 2023-09-29
Add the possibility to ignore "Type checking" imports

yakMM:
Closes https://github.com/seddonym/grimp/issues/64

~Sorry for the noise in the diff file, I had to change the order of classes and github is lost~

Task list:
- [x] Feature implementation for the import scanner
- [x] Import scanner documentation
- [x] Import scanner tests
- [x] Feature implementation for the cache
- [x] Cache documentation
- [x] Cache tests

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -17,4 +17,3 @@ omit = *migrations*
 exclude_lines =
     pragma: nocover
     @abc.abstractmethod
-    @overload
--- a/.coveragerc
+++ b/.coveragerc
@@ -17,3 +17,4 @@ omit = *migrations*
 exclude_lines =
     pragma: nocover
     @abc.abstractmethod
+    @overload
--- a/src/grimp/adaptors/caching.py
+++ b/src/grimp/adaptors/caching.py
@@ -34,7 +34,9 @@ class CacheFileNamer:
         # Use a hash algorithm with a limited size to avoid cache filenames that are too long
         # the filesystem, which can happen if there are more than a few root packages
         # being analyzed.
-        safe_unicode_identifier = hashlib.blake2b(bytes_identifier, digest_size=20).hexdigest()
+        safe_unicode_identifier = hashlib.blake2b(
+            bytes_identifier, digest_size=20
+        ).hexdigest()
         return f"{safe_unicode_identifier}.data.json"
 
     @classmethod
@@ -51,7 +53,9 @@ class CacheFileNamer:
         """
         package_names = (p.name for p in found_packages)
         csv_packages = ",".join(sorted(package_names))
-        include_external_packages_option = ":external" if include_external_packages else ""
+        include_external_packages_option = (
+            ":external" if include_external_packages else ""
+        )
         exclude_type_checking_imports_option = (
             ":no_type_checking" if exclude_type_checking_imports else ""
         )
--- a/src/grimp/adaptors/caching.py
+++ b/src/grimp/adaptors/caching.py
@@ -212,7 +212,9 @@ class Cache(AbstractCache):
             logger.warning(f"Could not use corrupt cache file {data_cache_filename}.")
             return {}
 
-        primitives_map: PrimitiveFormat = self._to_primitives_data_map(deserialized_json)
+        primitives_map: PrimitiveFormat = self._to_primitives_data_map(
+            deserialized_json
+        )
 
         return {
             Module(name=name): {
--- a/src/grimp/adaptors/caching.py
+++ b/src/grimp/adaptors/caching.py
@@ -21,24 +21,28 @@ class CacheFileNamer:
 
     @classmethod
     def make_data_file_name(
-        cls, found_packages: Set[FoundPackage], include_external_packages: bool
+        cls,
+        found_packages: Set[FoundPackage],
+        include_external_packages: bool,
+        exclude_type_checking_imports: bool,
     ) -> str:
         identifier = cls.make_data_file_unique_string(
-            found_packages, include_external_packages
+            found_packages, include_external_packages, exclude_type_checking_imports
         )
 
         bytes_identifier = identifier.encode()
         # Use a hash algorithm with a limited size to avoid cache filenames that are too long
         # the filesystem, which can happen if there are more than a few root packages
         # being analyzed.
-        safe_unicode_identifier = hashlib.blake2b(
-            bytes_identifier, digest_size=20
-        ).hexdigest()
+        safe_unicode_identifier = hashlib.blake2b(bytes_identifier, digest_size=20).hexdigest()
         return f"{safe_unicode_identifier}.data.json"
 
     @classmethod
     def make_data_file_unique_string(
-        cls, found_packages: Set[FoundPackage], include_external_packages: bool
+        cls,
+        found_packages: Set[FoundPackage],
+        include_external_packages: bool,
+        exclude_type_checking_imports: bool,
     ) -> str:
         """
         Construct a unique string that identifies the analysis parameters.
@@ -47,10 +51,13 @@ class CacheFileNamer:
         """
         package_names = (p.name for p in found_packages)
         csv_packages = ",".join(sorted(package_names))
-        include_external_packages_option = (
-            ":external" if include_external_packages else ""
+        include_external_packages_option = ":external" if include_external_packages else ""
+        exclude_type_checking_imports_option = (
+            ":no_type_checking" if exclude_type_checking_imports else ""
         )
-        return csv_packages + include_external_packages_option
+        return (
+            csv_packages + include_external_packages_option + exclude_type_checking_imports_option
+        )
 
 
 class Cache(AbstractCache):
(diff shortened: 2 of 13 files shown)


src/grimp/adaptors/importscanner.py
```diff
@@ -14,151 +15,28 @@ class NotAnImport(Exception):
     pass
 
 
-class ImportScanner(AbstractImportScanner):
```
seddonym, line 17: What's the reason for moving this down to the bottom of the module?

As `ImportScanner` is the public facing part of the module, I'd like it to stay up the top. I tried moving it to the top and everything seemed to work fine locally, but if you run into issues we could consider adding `from __future__ import annotations` which might help.

Separate reason: it's difficult to review this PR in its current form so I'd like to be able to understand what has changed better.

src/grimp/application/ports/importscanner.py
```diff
@@ -16,18 +16,21 @@ def __init__(
         file_system: AbstractFileSystem,
         found_packages: Set[FoundPackage],
         include_external_packages: bool = False,
+        include_type_checking_imports: bool = True,
```
seddonym, line 19: Could we change this to `exclude_type_checking_imports: bool = False` - and everywhere else? I'd like it to feel more like exceptional / unusual behaviour.

Also, non-blocking but I think this would be better as an argument passed to `scan_for_imports` rather than to `__init__`, as it makes the ImportScanner a little less stateful (we could use the same `ImportScanner` instead to scan in both ways).

src/grimp/application/usecases.py
```diff
@@ -91,23 +94,20 @@ def _validate_package_names_are_strings(
...
-            )
+            raise TypeError(f"Package names must be strings, got {name.__class__.__name__}.")
     return cast(Sequence[str], package_names)
 
 
 def _scan_packages(
     found_packages: Set[FoundPackage],
     file_system: AbstractFileSystem,
     include_external_packages: bool,
+    include_type_checking_imports: bool,
```
seddonym, line 105: We're going to need to think about caching here. We'll need a separate cache file based on whether the type checking import flag is passed. This shouldn't be too difficult - we already do it with the `include_external_packages` flag and I think this should work in the same way. It's important that existing caches (that don't exclude type checking imports) aren't invalidated, but I think this should just be a case of following the same approach as with `include_external_packages`.

More information about how caching currently works: https://grimp.readthedocs.io/en/stable/caching.html

src/grimp/application/usecases.py
```diff
@@ -30,6 +31,7 @@ def build_graph(
         - package_name: the name of the top level package for which to build the graph.
         - additional_package_names: tuple of the
         - include_external_packages: whether to include any external packages in the graph.
+        - include_type_checking_imports: whether to include imports made in type checking guards.
```
seddonym, line 34: We need to add this to the docs too. I'm happy to do this if you don't feel confident.

seddonym (changes requested):
Thanks for your hard work so far!

I haven't gone over the changes to `adaptdors/importscanner.py` that closely yet as it's tricky to review the diff, as the `ImportScanner` has been moved. Let's find a way of moving that back so it's clearer and then I'll look in more detail there.

We're also missing tests. I'd like as a minimum:

- Thorough test coverage of the new import scanner behaviour (in `tests/unit/adapters/test_importscanner.py`.
- At least one functional test which calls `build_graph` excluding type checking imports (and ideally that demonstrates that there is a difference in behaviour depending on whether you pass that argument).
- Test coverage for the cache naming strategy (in tests.unit.adaptors.test_caching.TestCache).

Any questions, just ask!

I have run this locally on a large graph and the graph builds about 15% faster than what we have currently, which is nice. 👏🏻

src/grimp/adaptors/importscanner.py
yakMM: The only reason behind moving it was to avoid the furture import. I'll move it back if it's okay

src/grimp/application/ports/importscanner.py
yakMM: I used exclude to match the other parameters, but hesitated a lot. Will change

src/grimp/application/usecases.py
yakMM: Yes if you look at the checkboxes in the PR presentation, I am aware of missing the cahce implementation for now

src/grimp/application/usecases.py
yakMM: Yep I still need to update docs and test, see the checkboxes as well

src/grimp/adaptors/importscanner.py
yakMM: Fixed, the diff should be better now

src/grimp/application/ports/importscanner.py
yakMM: Done

src/grimp/application/usecases.py
yakMM: Updated the doc for build_graph

src/grimp/application/usecases.py
yakMM: Caching should be fine after the last commits

yakMM:
@seddonym Added some doc and caching and updated the code according to your first review.

If I'm not mistaken only the tests for the new import scanner and build graph features are missing.

tests/unit/adaptors/test_caching.py
```diff
@@ -67,18 +86,23 @@ def test_make_data_file_name(self, include_external_packages, expected):
...
         )
 
         assert result == expected
 
     @pytest.mark.parametrize(
-        "include_external_packages, expected",
+        "include_external_packages, exclude_type_checking_imports, expected",
         (
-            (False, "hyphenated-package,underscore_package"),
-            (True, "hyphenated-package,underscore_package:external"),
```
yakMM, line 78: Got lazy to combine different parametrize together, I believe covering all the combinations manually is fine for now

yakMM:
Ready on my side, tell me if something is missing or if I should write deeper tests

tests/unit/adaptors/test_importscanner.py
```diff
@@ -743,6 +743,102 @@ def test_scans_multiple_packages(statement):
...
+    (
+        ("if t.TYPE_CHECKING:", True),
+        ("if TYPE_CHECKING:", True),
+        ("if typing.TYPE_CHECKING:", True),
+        ("if WEIRD_ALIAS.TYPE_CHECKING:", True),
+        ("if type_checking:", False),
+        ("while TYPE_CHECKING:", False),
+    ),
+)
+def test_exclude_type_checking_imports(
```
seddonym, line 758: Great test.

src/grimp/adaptors/caching.py
```diff
@@ -201,9 +212,7 @@ def _read_data_map_file(self) -> Dict[Module, Set[DirectImport]]:
             logger.warning(f"Could not use corrupt cache file {data_cache_filename}.")
             return {}
 
-        primitives_map: PrimitiveFormat = self._to_primitives_data_map(
-            deserialized_json
-        )
+        primitives_map: PrimitiveFormat = self._to_primitives_data_map(deserialized_json)
```
seddonym, line 215: There are lots of examples of changes to the formatting, which is presumably due to a mismatch between your editor line length settings and mine. I'd prefer not to include these formatting changes in this PR unless there is a good reason.

I thought [Black](https://pypi.org/project/black/) was already set up in CI but turns out it isn't. I've made [a ticket](https://github.com/seddonym/grimp/issues/131) to look into this which I'll look at in the next few days.

seddonym (changes requested):
This is a lovely piece of work, thank you!

I just have a couple of niggles: the addition of the `@overload`, and the changes to the formatting of some of the lines. I'll put in place something so that the formatting can be addressed easily (i.e. by running Black on it with the right line length) so this is caught by the checks.

Otherwise though this is good to go. Thanks again!

.coveragerc
```diff
@@ -17,3 +17,4 @@ omit = *migrations*
 exclude_lines =
     pragma: nocover
     @abc.abstractmethod
+    @overload
```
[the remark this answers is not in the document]
yakMM: The two overloads I added are for type checking only and will never be run, so they show as "uncovered" code without this line

yakMM:
Thanks for the review, I'd love to have black added to tox. I can remove the `@overload` if you don't like them, it's mostly used to make everything explicit (usually used in strongly typed context).

Here it's used to make the type of the ASTnode and the type of the node parser match (`Import` node and `ImportNodeParser` or `ImportFrom` node and `ImportFromNodeParser`, but no mix 

It's really not essential, especially since it's internal implem

.coveragerc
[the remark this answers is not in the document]
yakMM: Removed because I removed the `@overload`

yakMM:
Removed the `@overload` and trimmed out the formatting noise

seddonym (changes requested):
This is looking great.

There are git conflicts, probably because I just merged Black formatting. Probably if you run Black on your branch and then rebase it should be straightforward to resolve the conflicts. The line-length of 99 is already in `pyproject.toml` so it should just fall into place.

FYI Once that's done I'm going to merge your PR into a `type-checking` branch just to make a couple of tweaks to the docs before merging to master. That branch is the same as `master` currently.

Let me know if any concerns, and thanks again!

seddonym (approved):

PR #136, opened 2023-11-21
Allow layer sibling independence to be configured

Peter554:
This PR contains the changes needed within grimp to support [importlinter: Proposal: Allow configuring sibling independence](https://github.com/seddonym/import-linter/issues/208).

* The rust `find_illegal_dependencies` function is adapted to accept a tuple of level dicts, rather than a tuple of level sets. Each level dict contains the `layers` for that level (in a set) and the level dict also contains a boolean field `independent` which `find_illegal_dependencies` uses to determine whether the siblings within that layer should be independent.
* The python `graph.find_illegal_dependencies_for_layers` method is extended to accept a `Sequence[Union[str, set[str], Level]]`, where `Level` is a `TypedDict` matching the level dict expected by the rust `find_illegal_dependencies` function. 
* Existing behaviour is maintained - no breaking changes. A simple set of strings can still be passed to the python  `graph.find_illegal_dependencies_for_layers` API, and these sibling layers will be treated as independent, as before.


rust/src/lib.rs
```diff
@@ -47,15 +47,18 @@ pub fn find_illegal_dependencies<'a>(
     convert_dependencies_to_python(py, dependencies, &graph)
 }
 
-fn rustify_levels(levels_python: &PyTuple) -> Vec<Level> {
+fn rustify_levels(levels_python: &PyTuple) -> PyResult<Vec<Level>> {
     let mut rust_levels: Vec<Level> = vec![];
     for level_python in levels_python.into_iter() {
-        let layers: HashSet<&str> = level_python.extract().unwrap();
+        let level_dict = level_python.downcast::<PyDict>()?;
+        let independent: bool = level_dict.get_item("independent").unwrap().extract()?;
```
Peter554, line 54: `unwrap`ing the `Option` can panic here if the key is not defined. Since this rust crate is only used internally within grimp, and we know for certain that grimp is passing this key then I think this is fine.

src/grimp/adaptors/_layers.py
```diff
@@ -175,193 +174,3 @@ def _module_from_layer(layer: str, container: str | None = None) -> _Module:
     else:
         name = layer
     return _Module(name)
-
-
-def _search_for_package_dependency(
```
Peter554, line 166: Removing this dead code is of course not really related to the topic of this PR, just cleaning up. I assume there is no reason that we should be keeping this code anymore?

Peter554:
@seddonym this is ready for review now. Hopefully it's along the right lines, if not then let's just have a quick chat.

seddonym:
Thank you! Will take a look over the next few days.

Peter554:
> Thank you! Will take a look over the next few days.

Thanks! PR seems to be closed though, was that intentional?

seddonym:
Oops no not at all, sorry! Will reopen when at a computer

src/grimp/adaptors/_layers.py
```diff
@@ -52,11 +53,24 @@ class _RustPackageDependency(TypedDict):
     routes: tuple[_RustRoute, ...]
 
 
-def _layers_to_levels(layers: Sequence[Union[str, set[str]]]) -> tuple[set[str], ...]:
+def _layers_to_levels(layers: Sequence[Union[str, set[str], Level]]) -> tuple[Level, ...]:
```
seddonym, line 55: This change to the API feels quite significant, and is the only thing that concerns me about this PR. As you've probably noticed, up until now the API just uses native Python datatypes and I think that has worked well generally, though it is being stretched a little with the more complex return types of `find_illegal_dependencies_for_layers` and `get_import_details`. I do really like the way we haven't needed any custom datatypes so far, but I'm starting to wonder if it's time to move beyond this as the API grows in complexity.

For me, passing it in as a dict is `dict(independent=False, layers={"green", "blue"})` probably crosses a boundary: a custom datatype would be simpler here. That said, I do really like the approach of backward compatibility you've taken. I've been trying to think of alternatives.

One option (let's call it Option A) would be to define `Level` as a custom class (not just a `TypedDict`). The API might look like this:

```
dependencies = graph.find_illegal_dependencies_for_layers(
    layers=(
        "red",
        grimp.Level({"green", "blue"}, independent=False),
        "yellow",
    ),
)
```

Notice `independent` is moved to the end and the layers have been made a positional argument, for terseness.

Another even terser option (Option B) would be to make all the layers positional so we don't need to pass them in as a set, i.e. `grimp.Level("green", "blue", independent=False)`.

With either of these options, I think it would be nice to make `Level` the canonical datatype passed to the function, with the `str` and `set[str]` options being conveniences that are converted to `Level` objects straight away. So we could also call the function like this:

```
dependencies = graph.find_illegal_dependencies_for_layers(
    layers=(
        grimp.Level("red"),
        grimp.Level("green", "blue", independent=False),
        grimp.Level("yellow", "orange"),
    ),
)
```

If we did that then it would make sense to annotate `layers` as `Level | str | set[str]`, to make level more prominent. Or possibly we should even rename it to `levels` (we'd need to figure out backward compatibility though).

Another option (Option C) I considered was to use the decorator pattern just to wrap the set:

```
dependencies = graph.find_illegal_dependencies_for_layers(
    layers=(
        "red",
        grimp.dependent({"green", "blue"}),
        "yellow",
    ),
)
```

This has the advantage of being very terse, though I do wonder if it is potentially confusing.

I'm leaning towards Option B, interested to hear your thoughts.

Peter554: I was originally thinking to introduce a class for this, a bit like you're suggesting. I had decided I wouldn't mainly because of the internal rust function - when calling the rust function it seemed easiest to work with a primative dict rather than a more complex type. Maybe there is a way to pass complex types to rust via `pyo3` though (do you happen to know?). Anyways, this internal decision shouldn't affect the public API of the python package, I agree.

I think option B works for me. 

I wonder about the naming - is `grimp.Level` meaningful enough? Or should it be something more verbose but explicit e.g. `grimp.LayeredArchitectureLevel`? What do you think?

Peter554: Should this `Level` type (or whatever we decide to name it) be a `ValueObject` https://github.com/seddonym/grimp/blob/master/src/grimp/domain/valueobjects.py?

seddonym: > when calling the rust function it seemed easiest to work with a primative dict rather than a more complex type

Agreed. It's fine to convert the object to a dict in Python before passing to Rust, that's just an internal implementation detail that can be easily changed.

rust/src/layers.rs
```diff
@@ -69,13 +69,18 @@ fn _generate_module_permutations<'a>(
                     continue;
                 }
 
-                // Build the layers that mustn't import this higher layer. That
-                // includes lower layers and siblings.
+                // Build the layers that mustn't import this higher layer.
```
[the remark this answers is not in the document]
Peter554: Yes added (used `///` else the IDE won't recognize this as a docstring, `//` is just a comment).

rust/src/lib.rs
```diff
@@ -47,15 +47,18 @@ pub fn find_illegal_dependencies<'a>(
     convert_dependencies_to_python(py, dependencies, &graph)
 }
 
-fn rustify_levels(levels_python: &PyTuple) -> Vec<Level> {
+fn rustify_levels(levels_python: &PyTuple) -> PyResult<Vec<Level>> {
```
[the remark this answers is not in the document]
Peter554: Updated.

tests/unit/adaptors/graph/test_layers.py
```diff
@@ -435,6 +438,24 @@ def test_direct_illegal_between_sibling_layers(
             ),
         }
 
+    def test_imports_between_sibling_layers_illegal_if_and_only_if_level_is_independent(self):
```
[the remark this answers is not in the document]
Peter554: Sure, I've extended the existing `test_direct_illegal_between_sibling_layers` now.

src/grimp/application/ports/graph.py
```diff
@@ -15,6 +15,11 @@ class DetailedImport(TypedDict):
     line_contents: str
 
 
+class Level(TypedDict):
```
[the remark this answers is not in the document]
Peter554: Yes, added.

src/grimp/adaptors/_layers.py
Peter554: Updated in the option B fashion 🤞

docs/usage.rst
```diff
@@ -314,34 +329,47 @@ Higher level analysis
...
+
+    .. attribute:: independent
+
+    ``bool``: Whether the sibling layers within this level are required to be independent.
+
+.. class:: PackageDependency
+
+    A collection of import dependencies from one Python package to another.
 
-      .. attribute:: importer
```
Peter554, line 321: There were some indentation issues here, which got fixed here too.

Before

Screenshot 2023-12-14 at 19 17 05

After

Screenshot 2023-12-14 at 19 17 09

src/grimp/application/ports/graph.py
```diff
@@ -284,13 +284,15 @@ def find_illegal_dependencies_for_layers(
 
         Additionally, multiple layers can be grouped together at the same level; for example
         `mypackage.utils` and `mypackage.logging` might sit at the bottom, so they cannot
-        import from any other layers. Layers at the same level must be independent, so any
-        dependencies in either direction will be treated as illegal.
+        import from any other layers. When a simple set of sibling layers is passed then they
```
seddonym, line 287: Would slightly prefer the documentation (at least in this docstring) to indicate `Level` as being the primary type that is passed, and then mention the single-string / set of strings as conveniences that will be converted to levels. But not a blocker.

seddonym (approved):
Really good stuff! Thanks for such a high quality pull request.

PR #137, opened 2024-01-03
Rename Level to Layer

Peter554:
Following discussions with @seddonym. A suggested renaming:

* `grimp.Level` to `grimp.Layer` - since a "level" is actually a layer (these terms are currently mixed and used to mean the same thing, lets just use one terminology).
* `grimp.Level.layers` to `grimp.Layer.module_tails` - since there are no "layers" _within_ a layer/level, the items within a layer are actually modules (or the "tails" of modules, in the case of `containers`).

Right now this is just the python changes - we should probably align the rust code too, although for this PR these python changes might be enough, so that we can prioritise tidying the package public API (so it is publishable) and also unblock https://github.com/seddonym/import-linter/pull/209.

### DIFF
--- a/src/grimp/domain/valueobjects.py
+++ b/src/grimp/domain/valueobjects.py
@@ -86,21 +86,21 @@ class DirectImport(ValueObject):
         return hash((str(self), self.line_contents))
 
 
-class Level(ValueObject):
+class Layer(ValueObject):
     """
-    A level within a layered architecture.
+    A layer within a layered architecture.
 
-    If level.independent is True then the layers within the level are considered
+    If layer.independent is True then the modules within the layer are considered
     independent. This is the default.
     """
 
     def __init__(
         self,
-        *layers: str,
+        *module_tails: str,
         independent: bool = True,
     ) -> None:
-        self.layers = set(layers)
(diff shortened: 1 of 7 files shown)


src/grimp/domain/valueobjects.py
```diff
@@ -86,21 +86,21 @@ def __hash__(self) -> int:
...
 
-    If level.independent is True then the layers within the level are considered
+    If layer.independent is True then the modules within the layer are considered
     independent. This is the default.
     """
 
     def __init__(
         self,
-        *layers: str,
+        *module_tails: str,
```
Peter554, line 99: 🤔 Personally I think I still slightly prefer to call this simply `modules` or `module_names`. I know it's not _technically_ correct when `containers` are used, but I think overall it's probably still a bit simpler to understand. Not a strong opinion though.

seddonym: Let's stick with module tails. I think it's valuable to have precise language in the guts of Grimp. I don't think users of Grimp will have to engage much with this terminology in practice.

seddonym:
Thanks so much Peter! 👏🏻

PR #142, opened 2024-02-12
Speed up import scanning by 400-500%

duzumaki:
Tested on a large mono repo (35,000 modules) using a mac m1 (8 cores). The modules are divided amongst the cores on the system it runs.
so each core is doing an even share of ast walking 

profiled using `Cprofile`, `pstats`(to read the profile dump) and `snakeviz`(for the visualisation)

Current(~37s):
Screenshot 2024-02-12 at 00 50 30

after parallelisation(~8s):
Screenshot 2024-02-12 at 00 52 35

### DIFF
--- a/src/grimp/application/usecases.py
+++ b/src/grimp/application/usecases.py
@@ -1,9 +1,13 @@
 """
 Use cases handle application logic.
 """
+
+from concurrent.futures import ProcessPoolExecutor
+from functools import partial
+from math import ceil
 from typing import Dict, Sequence, Set, Type, Union, cast
 
-from ..application.ports import caching
+from ..application.ports import caching, modulefinder
 from ..application.ports.filesystem import AbstractFileSystem
 from ..application.ports.graph import ImportGraph
 from ..application.ports.importscanner import AbstractImportScanner
@@ -99,6 +103,28 @@ def _validate_package_names_are_strings(
     return cast(Sequence[str], package_names)
 
 
+def get_imports_by_module(
+    module_file: modulefinder.ModuleFile,
+    cache_dir: Union[str, Type[NotSupplied], None],
+    import_scanner: AbstractImportScanner,
+    exclude_type_checking_imports: bool,
+    cache: caching.Cache,
+):
+    module = module_file.module
+    direct_imports = set()
+
+    try:
+        if cache_dir is None:
+            raise caching.CacheMiss
+        direct_imports = cache.read_imports(module_file)
+    except caching.CacheMiss:
+        direct_imports = import_scanner.scan_for_imports(
+            module, exclude_type_checking_imports=exclude_type_checking_imports
+        )
+
+    return module, direct_imports
+
+
 def _scan_packages(
     found_packages: Set[FoundPackage],
     file_system: AbstractFileSystem,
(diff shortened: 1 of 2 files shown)


tests/functional/test_error_handling.py
```diff
@@ -1,25 +1,15 @@
...
 
-def test_syntax_error_includes_module():
-    dirname = os.path.dirname(__file__)
-    filename = os.path.abspath(
-        os.path.join(dirname, "..", "assets", "syntaxerrorpackage", "foo", "one.py")
-    )
-
-    with pytest.raises(exceptions.SourceSyntaxError) as excinfo:
+def test_syntax_error_terminates_executor_pool():
+    with pytest.raises(BrokenProcessPool):
```
duzumaki, line 10: I guess this is one disadvantage of concurrency in python. using the .map() method in the executor results in the inability to capture errors raised from individual tasks

seddonym: It would be a shame to lose that exception - returning any syntax errors (they could even be exception objects) sounds like it is at least worth a try.

src/grimp/application/usecases.py
```diff
@@ -122,18 +150,25 @@ def _scan_packages(
...
-                if cache_dir is None:
-                    raise caching.CacheMiss
-                direct_imports = cache.read_imports(module_file)
-            except caching.CacheMiss:
-                direct_imports = import_scanner.scan_for_imports(
-                    module, exclude_type_checking_imports=exclude_type_checking_imports
-                )
-            imports_by_module[module] = direct_imports
+        with ProcessPoolExecutor() as executor:
+            chunk_size = ceil(len(found_package.module_files) / executor._max_workers) or 1
```
seddonym, line 163: Would be helpful to have a comment here explaining our reasoning for the chunk size.

seddonym:
Very exciting!

Have left a couple of initial comments, also the tests are failing which would be good to sort out.

src/grimp/application/usecases.py
```diff
@@ -99,6 +103,28 @@ def _validate_package_names_are_strings(
     return cast(Sequence[str], package_names)
 
 
+def get_imports_by_module(
+    module_file: modulefinder.ModuleFile,
+    cache_dir: Union[str, Type[NotSupplied], None],
+    import_scanner: AbstractImportScanner,
+    exclude_type_checking_imports: bool,
+    cache: caching.Cache,
+):
```
seddonym, line 112: Missing type annotation on return value.

wesleykendall:
I checked out this branch and installed it locally. It dramatically slowed down graph building time on a large test repo (0.6 seconds -> 35 seconds).

On a Mac M2. I verified that my local installation of the `master` branch on this repo yielded fast results, so I don't think I'm doing anything wrong on the installation.

Seems like something is wrong. When I set `max_workers` to `1`, build time was around 10 seconds. `max_workers` of `2` slowed down to 14 seconds. Let me know if there is another way I can profile

seddonym:
I'm going to close this branch because we are close to having a Rust implementation of the graph, and so I think the next step will be to move to Rust for building the graph. It'll be easier to parallelise there because we can free ourselves from the GIL and do multithreading.

But thanks for your efforts, much appreciated!

PR #152, opened 2024-06-12
module files: discover files in symlinked folders

sshane:
I am not too familiar with this module, just trying to fix an issue I came across when trying to use another tool (linked below). If there are any reasons not to do this please let me know so I can take another look! We can also make this a parameter if that's preferred.


seddonym (changes requested):
Thanks for doing this!

Before I merge, would you mind adding a line to the CHANGELOG and (if you like) add your name to `AUTHORS.rst`?

sshane:
done!

seddonym (approved):
Thanks!

First time I've come across symlinked python packages, but I don't see why not.

PR #157, opened 2024-07-05
Add as_packages parameter to find_shortest_chains

NicholasBunn:
Adds the 'as_packages' option to the 'find_shortest_chains' methods. This allows the importer to be treated as a module rather than a package. The default behaviour is to treat it as a package so as to maintain backwards compatibility

### DIFF
--- a/src/grimp/adaptors/graph.py
+++ b/src/grimp/adaptors/graph.py
@@ -304,8 +304,12 @@ class ImportGraph(graph.ImportGraph):
         """
         shortest_chains = set()
 
-        upstream_modules = self._all_modules_in_package(imported)
-        downstream_modules = self._all_modules_in_package(importer)
+        upstream_modules = (
+            {imported} if not as_packages else self._all_modules_in_package(imported)
+        )
+        downstream_modules = (
+            {importer} if not as_packages else self._all_modules_in_package(importer)
+        )
 
         if upstream_modules & downstream_modules:
             # If there are shared modules between the two, one of the modules is a descendant
--- a/src/grimp/adaptors/graph.py
+++ b/src/grimp/adaptors/graph.py
@@ -287,12 +287,17 @@ class ImportGraph(graph.ImportGraph):
 
         return self._find_shortest_chain(importer=importer, imported=imported)
 
-    def find_shortest_chains(self, importer: str, imported: str) -> Set[Tuple[str, ...]]:
+    def find_shortest_chains(
+        self, importer: str, imported: str, as_packages: bool = True
+    ) -> Set[Tuple[str, ...]]:
         """
         Find the shortest import chains that exist between the importer and imported, and
-        between any modules contained within them. Only one chain per upstream/downstream pair
-        will be included. Any chains that are contained within other chains in the result set
-        will be excluded.
+        between any modules contained within them if as_packages is True. Only one chain per
+        upstream/downstream pair will be included. Any chains that are contained within other
+        chains in the result set will be excluded.
+
+        The default behavior is to treat the import and imported as packages, however, if
+        as_packages is False, both the importer and imported will be treated as modules instead.
 
         Returns:
             A set of tuples of strings. Each tuple is ordered from importer to imported modules.
--- a/tests/functional/test_build_and_use_graph.py
+++ b/tests/functional/test_build_and_use_graph.py
@@ -1,4 +1,6 @@
 from grimp import build_graph
+from typing import Set, Tuple, Optional
+import pytest
 
 """
 For ease of reference, these are the imports of all the files:
@@ -15,6 +17,12 @@ testpackage.two.alpha: testpackage.one.alpha
 testpackage.two.beta: testpackage.one.alpha
 testpackage.two.gamma: testpackage.two.beta, testpackage.utils
 testpackage.utils: testpackage.one, testpackage.two.alpha
+testpackage.three: testpackage.one.alpha
+testpackage.three.alpha: testpackage.one.alpha
+testpackage.three.beta: testpackage.one.alpha
+testpackage.three.gamma: testpackage.two.beta, testpackage.utils
+
+
 
 """
 
@@ -38,6 +46,10 @@ def test_modules():
         "testpackage.two.beta",
         "testpackage.two.gamma",
         "testpackage.utils",
+        "testpackage.three",
+        "testpackage.three.beta",
+        "testpackage.three.gamma",
+        "testpackage.three.alpha",
     }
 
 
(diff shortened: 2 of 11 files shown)


tests/unit/adaptors/graph/test_chains.py
```diff
@@ -234,7 +234,24 @@ def test_short_indirect_import(self):
 
         assert result == {("green.indirect", "purple", "blue.foo")}
 
-    def test_long_indirect_import(self):
+    def test_short_indirect_import_as_module(self):
```
seddonym, line 237: I feel we could do with a bit more test coverage in general. Maybe each of the test cases in `TestFindShortestChains` should be parametrized with `as_packages=True/False` and then make sure the graph is constructed so we get different expected in each case. What do you think?

tests/unit/adaptors/graph/test_chains.py
```diff
@@ -234,7 +234,24 @@ def test_short_indirect_import(self):
...
+        graph = ImportGraph()
+        graph.add_module("green")
```
seddonym, line 239: Feels like we're adding a lot of modules unnecessarily in this test case - compare with `test_short_indirect_import_as_packages`?

src/grimp/adaptors/graph.py
```diff
@@ -301,8 +301,12 @@ def find_shortest_chains(
         """
         shortest_chains = set()
 
-        upstream_modules = self._all_modules_in_package(imported)
-        downstream_modules = self._all_modules_in_package(importer)
+        upstream_modules = (
+            {imported} if not as_packages else self._all_modules_in_package(imported)
```
seddonym, line 305: Simple!

seddonym (changes requested):
This is great stuff, thanks for taking a look!

I think it would probably be sensible to add a bit more test coverage - but happy to take a look.

tests/unit/adaptors/graph/test_chains.py
NicholasBunn: Yea, I've added some cases that are purposefully a bit longer to ensure we're getting the shortest ones we can. But I see this isn't the way other tests are done, can update quickly 👍🏼

NicholasBunn: I've updated it so that we have one import of the blue module (green-purple-blue), and a shorter import of a module in the blue package (green-blue.foo) to ensure that we get the shortest path to a module if `as_packages` is False

tests/unit/adaptors/graph/test_chains.py
NicholasBunn: Yea I think that sounds good! Adding it in 👍🏼

tests/functional/test_build_and_use_graph.py
```diff
@@ -199,6 +199,17 @@ def test_find_shortest_chains():
     } == graph.find_shortest_chains(importer="testpackage.two", imported="testpackage.one")
 
 
+def test_find_shortest_chains_as_module():
+    graph = build_graph("testpackage", cache_dir=None)
+    graph.add_import(importer="testpackage.two", imported="testpackage.one.alpha")
```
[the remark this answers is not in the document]
NicholasBunn: Added a third test package 👍🏼

tests/unit/adaptors/graph/test_chains.py
NicholasBunn: Alright I think I've parametrized everything that's relevant here - all of the tests bar the ones related to long chains

tests/unit/adaptors/graph/test_chains.py
```diff
@@ -167,85 +169,188 @@ def test_demonstrate_nondeterminism_of_equal_chains(self):
 
 
 class TestFindShortestChains:
-    def test_top_level_import(self):
+    @pytest.mark.parametrize(
+        "as_packages, expected_result",
+        (
+            (False, {("green", "blue")}),
```
seddonym, line 175: No need to have the expected result as a parameter if it's the same.

tests/unit/adaptors/graph/test_chains.py
```diff
@@ -167,85 +169,188 @@ def test_demonstrate_nondeterminism_of_equal_chains(self):
...
+        )
 
-        assert result == {("green.foo", "blue.bar")}
+        assert result == expected_result
 
-    def test_no_results_in_reverse_direction(self):
+    @pytest.mark.parametrize(
+        "as_packages, expected_result",
+        (
+            (False, set()),
```
seddonym, line 211: ♻️ No need to parametrize `expected_result`.

tests/unit/adaptors/graph/test_chains.py
```diff
@@ -327,27 +432,45 @@ def test_chains_that_reenter_imported_package_dont_stop_longer_chains_being_incl
             ),
         }
 
-    def test_doesnt_lose_import_details(self):
+    @pytest.mark.parametrize(
+        "as_packages",
+        (
+            False,
+            True,
```
seddonym, line 439: Could we remove any necessary commas here so the `parametrize` decorator all gets formatted onto one line?

seddonym (changes requested):
Looking good. I just have a few very small tweaks, if it's not too much trouble?

Additionally, could you:

- Add yourself to CONTRIBUTORS.md
- Add a line to the CHANGELOG (put it under a `latest` section at the top).

Thanks again!

tests/functional/test_build_and_use_graph.py
```diff
@@ -189,14 +204,62 @@ def test_find_shortest_chain():
...
+                ("testpackage.three.alpha", "testpackage.one.alpha"),
+                ("testpackage.three.beta", "testpackage.one.alpha"),
+                (
+                    "testpackage.three.gamma",
+                    "testpackage.utils",
+                    "testpackage.two.alpha",
+                    "testpackage.one.alpha",
+                ),
+            },
+            id="Treating imports as modules (implicit)",
```
seddonym, line 245: "modules" should be "packages". I can change this though.

tests/unit/adaptors/graph/test_chains.py
```diff
@@ -167,85 +169,188 @@ def test_demonstrate_nondeterminism_of_equal_chains(self):
...
-        result = graph.find_shortest_chains(importer="green", imported="blue")
+        result = graph.find_shortest_chains(
+            importer="green", imported="blue", as_packages=as_packages
+        )
 
-        assert result == {("green.foo", "blue.bar")}
+        assert result == expected_result
 
-    def test_no_results_in_reverse_direction(self):
+    @pytest.mark.parametrize(
```
seddonym, line 208: FYI We can do this whole block as

```
@pytest.mark.parametrize("as_packages", (True, False))
```

seddonym (approved):

seddonym:
Great stuff! Thanks for all your help.

PR #166, opened 2024-11-01
Rust graph

seddonym:
Implements the graph in Rust. 🥳 

It took a while to get something that performs better - thanks @Peter554 for figuring out the final implementation.


rust/src/graph/higher_order_queries.rs
```diff
@@ -0,0 +1,210 @@
...
+
+        // Disallow chains via these imports.
+        // We'll add chains to this set as we discover them.
+        let mut excluded_imports = FxHashMap::default();
+
+        // Collect direct imports...
+        let mut direct_imports = vec![];
+        // ...and the middles of any indirect imports.
+        let mut middles = vec![];
+        loop {
```
seddonym, line 138: How confident are we that this loop will complete? I don't think there are any tests for handling cycles in `find_illegal_dependencies_for_layers`, we should probably add some.

src/grimp/adaptors/graph.py
```diff
@@ -129,344 +52,149 @@ def add_import(
...
     def find_modules_that_directly_import(self, module: str) -> Set[str]:
-        return self._importers_by_imported[module]
-
-    def get_import_details(self, *, importer: str, imported: str) -> List[graph.DetailedImport]:
-        import_details_for_importer = self._import_details.get(importer, [])
-        # Only include the details for the imported module.
-        # Note: we copy each details dictionary at this point, as our deepcopying
-        # only copies the dictionaries by reference.
-        return [i.copy() for i in import_details_for_importer if i["imported"] == imported]
+        if module in self._rustgraph.get_modules():
```
seddonym, line 93: This check is costly. Let's add a benchmark and then (probably) push the check down to Rust.

src/grimp/adaptors/graph.py
```diff
@@ -15,6 +15,7 @@ class ImportGraph(graph.ImportGraph):
 
     def __init__(self) -> None:
         super().__init__()
+        self._cached_modules: Set[str] | None = None
```
Peter554, line 18: This trick helps, but not as much as you currently think. We need to make sure that the property isn't already set when running the benchmarks. Try this diff

```diff
diff --git a/tests/benchmarking/test_benchmarking.py b/tests/benchmarking/test_benchmarking.py
index 99414da..1c4ea80 100644
--- a/tests/benchmarking/test_benchmarking.py
+++ b/tests/benchmarking/test_benchmarking.py
@@ -421,6 +421,7 @@ def test_copy_graph(large_graph, benchmark):

 def test_graph_contains_module(large_graph, benchmark):
     def f(n):
+        large_graph._cached_modules = None
         for i in range(n):
             _ = f"foo{i}" in large_graph.modules

@@ -429,6 +430,7 @@ def test_graph_contains_module(large_graph, benchmark):

 def test_iterate_over_modules_in_graph(large_graph, benchmark):
     def f():
+        large_graph._cached_modules = None
         for module in large_graph.modules:
             _ = module
```

Effect 👇 

```
-------- benchmark 'test_graph_contains_module': 2 tests ---------
Name (time in us)                                   Mean
------------------------------------------------------------------
test_graph_contains_module (0140_e4e7b39)        11.1467 (1.0)
test_graph_contains_module (NOW)              5,601.3647 (502.51)
------------------------------------------------------------------

-------- benchmark 'test_iterate_over_modules_in_graph': 2 tests ---------
Name (time in us)                                           Mean
--------------------------------------------------------------------------
test_iterate_over_modules_in_graph (0140_e4e7b39)       485.2748 (1.0)
test_iterate_over_modules_in_graph (NOW)              6,361.4360 (13.11)
--------------------------------------------------------------------------
```

rust/src/graph/higher_order_queries.rs
seddonym: We've chatted - I'm more confident now.

PR #173, opened 2025-01-16
Experiment with changes to the BiMap library

andir:


seddonym:
Thanks for this. As discussed I'm going to close as there have been a lot of changes to the implementation on the `rust-graph` branch which seem to have sorted the performance issues out, but this is helpful to have in our back pocket if we need to go further.

PR #175, opened 2025-01-17
Rust graph (Peter)

Peter554:
This PR contains a rewrite of the import graph in rust. A lot of the code in this PR is ported from https://github.com/Peter554/pyimports.

Performance tricks in this PR:
* I've restructured the graph. There are no strings anymore - the graph is just ints (and tokens, which are pairs of ints).
  * [slotmap](https://docs.rs/slotmap/) as the main data structure.
  * [string-interner](https://docs.rs/string-interner/) for strings.
* I've tried to reduce cloning as much as possible, especially of the entire graph.
* I've found some methods for more efficient pathfinding. 
  * Do one BFS with multiple start nodes rather than a for loop of many BFS. This is useful in e.g. `find_downstream_modules`, `chain_exists`.
  * The pathfinding algorithm allows us to dynamically exclude imports, without having to clone and mutate the entire graph. This is useful in e.g. `find_shortest_chains` and `find_illegal_dependencies_for_layers`. 
* I noticed that we were doing `if module in self.modules` in the python code, sometimes in a for loop. This is actually rather inefficient, since the entire graph has to be copied over from rust to a python set (every time). I added `graph.contains_module` as a more performant alternative. We should ideally find a way to avoid others making this same mistake. At the very least, document this. Perhaps also make it a method rather than a property e.g. `graph.modules()`, to show more clearly that work is being done (breaking change, but perhaps reasonable).
* I've short circuited the implementation of `find_shortest_chains` and `find_illegal_dependencies_for_layers` in the case that the contract is kept (via an initial `chain_exists` check). This should give significant benefits for users running import linter in CI, since most of the time most contracts should be kept - it makes sense to optimise for this case.
* Bidirectional shortest path algorithm.


rust/src/graph/mod.rs
```diff
@@ -0,0 +1,137 @@
...
+pub mod import_chain_queries;
+
+pub(crate) mod pathfinding;
+
+lazy_static! {
+    static ref MODULE_NAMES: RwLock<StringInterner<StringBackend>> =
+        RwLock::new(StringInterner::default());
+    static ref EMPTY_MODULE_TOKENS: FxHashSet<ModuleToken> = FxHashSet::default();
+    static ref EMPTY_IMPORT_DETAILS: FxHashSet<ImportDetails> = FxHashSet::default();
+    static ref EMPTY_IMPORTS: FxHashSet<(ModuleToken, ModuleToken)> = FxHashSet::default();
```
seddonym, line 24: 🤔 Rather than modelling imports as two-tuples of `ModuleToken`s, might be better to use a struct so it's more explicit which is the importer and which the imported. What do you think?

rust/src/graph/mod.rs
```diff
@@ -0,0 +1,137 @@
...
+}
+
+new_key_type! { pub struct ModuleToken; }
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Getters, CopyGetters)]
+pub struct Module {
+    #[getset(get_copy = "pub")]
+    token: ModuleToken,
```
seddonym, line 32: So is this just a way of stopping us needing to call `module.token.clone()` all over the place? Instead we just call `module.token()`?

rust/src/graph/mod.rs
```diff
@@ -0,0 +1,137 @@
...
+    // Invisible modules exist in the hierarchy but haven't been explicitly added to the graph.
+    #[getset(get_copy = "pub")]
+    is_invisible: bool,
+
+    #[getset(get_copy = "pub")]
+    is_squashed: bool,
+}
+
+impl Module {
+    pub fn name_as_string(&self) -> String {
```
seddonym, line 46: Might be more ergonomic to call this `name` and call the field `interned_name`, `name_symbol` or something like that?

rust/src/graph/mod.rs
```diff
@@ -0,0 +1,137 @@
...
+        let interner = MODULE_NAMES.read().unwrap();
```
seddonym, line 47: TODO: talk me through why the unwraps in this are safe.

rust/src/graph/mod.rs
```diff
@@ -0,0 +1,137 @@
...
+        interner.resolve(self.name).unwrap().to_owned()
+    }
+}
+
+#[derive(Default, Clone)]
+pub struct Graph {
+    // Hierarchy
+    modules_by_name: BiMap<DefaultSymbol, ModuleToken>,
+    modules: SlotMap<ModuleToken, Module>,
+    module_parents: SecondaryMap<ModuleToken, Option<ModuleToken>>,
```
seddonym, line 57: Ok so no Petgraph? Would be interesting to talk through.

rust/src/graph/mod.rs
```diff
@@ -0,0 +1,137 @@
...
+    pub fn name_as_string(&self) -> String {
+        let interner = MODULE_NAMES.read().unwrap();
+        interner.resolve(self.name).unwrap().to_owned()
+    }
+}
+
+#[derive(Default, Clone)]
+pub struct Graph {
+    // Hierarchy
+    modules_by_name: BiMap<DefaultSymbol, ModuleToken>,
```
seddonym, line 55: Let's talk through all the data structures here, we have quite a few different maps and it would be good to sense check each one.

rust/src/graph/mod.rs
```diff
@@ -0,0 +1,137 @@
...
+    }
+}
+
+impl From<ModuleToken> for FxHashSet<ModuleToken> {
+    fn from(value: ModuleToken) -> Self {
+        FxHashSet::from_iter([value])
+    }
+}
+
+pub trait ExtendWithDescendants:
```
seddonym, line 77: TODO talk me through this one.

rust/src/graph/mod.rs
```diff
@@ -0,0 +1,137 @@
...
+
+new_key_type! { pub struct ModuleToken; }
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Getters, CopyGetters)]
+pub struct Module {
+    #[getset(get_copy = "pub")]
+    token: ModuleToken,
+
+    #[getset(get_copy = "pub")]
+    name: DefaultSymbol,
```
seddonym, line 35: This, and the `token` field, are both effectively tokens right? I wonder if it is possible to combine them.

rust/src/graph/mod.rs
```diff
@@ -0,0 +1,137 @@
...
+pub trait ModuleTokenIterator<'a>: Iterator<Item = &'a ModuleToken> + Sized {
+    fn into_module_iterator(self, graph: &'a Graph) -> impl ModuleIterator<'a> {
+        self.map(|m| graph.get_module(*m).unwrap())
+    }
+}
+
+impl<'a, T: Iterator<Item = &'a ModuleToken>> ModuleTokenIterator<'a> for T {}
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash, new, Getters, CopyGetters)]
+pub struct ImportDetails {
```
seddonym, line 130: This is maybe using slightly different terminology to https://grimp.readthedocs.io/en/stable/usage.html#ImportGraph.get_import_details. I'm not massively happy with the name I chose for that method originally. But I think it might be worth coming up with a non-plural name for the struct here as it's more ergonomic. `ImportMetadata`?

rust/src/lib.rs
```diff
@@ -34,54 +48,66 @@ impl GraphWrapper {
         }
     }
 
-    pub fn get_modules(&self) -> FxHashSet<String> {
+    pub fn get_modules(&self) -> HashSet<String> {
```
seddonym, line 51: What's the reason for this change?

rust/src/lib.rs
```diff
@@ -34,54 +48,66 @@ impl GraphWrapper {
...
         self._graph
-            .get_modules()
-            .iter()
-            .map(|module| module.name.clone())
+            .all_modules()
```
seddonym, line 53: Any reason why we're renaming this method? Maybe easier to have the same name as on the wrapper?

rust/src/lib.rs
```diff
@@ -34,54 +48,66 @@ impl GraphWrapper {
...
     pub fn add_module(&mut self, module: &str, is_squashed: bool) -> PyResult<()> {
-        let module_struct = Module::new(module.to_string());
-
-        if let Some(ancestor_squashed_module) =
-            self._graph.find_ancestor_squashed_module(&module_struct)
+        for ancestor_module in self
+            ._graph
+            .module_name_to_self_and_ancestors(module)
+            .into_iter()
+            .skip(1)
```
seddonym, line 73: Could you explain this line?

seddonym (commented):
Leaving some initial comments but I've still got lots to read through and digest.

src/grimp/application/ports/graph.py
```diff
@@ -32,6 +32,13 @@ def modules(self) -> Set[str]:
         """
         raise NotImplementedError
 
+    @abc.abstractmethod
```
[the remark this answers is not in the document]
Peter554: As mentioned in slack, I think this might be a slightly trickier problem than we thought. I had a little play here https://github.com/Peter554/grimp/pull/1 but not real luck yet.

Hopefully we'll find a way to address this. If not from my perspective a breaking change + major version release is probably acceptable for such a big internal change (and the breaking change is not that hard to adapt to as a consumer). Let's see.

Anyways, let's leave it out of scope for the merge into `rust-graph` and worry about it more later.

rust/src/errors.rs
```diff
@@ -0,0 +1,29 @@
...
+    #[error("Container {0} does not exist.")]
+    NoSuchContainer(String),
+
+    #[error("Modules have shared descendants.")]
+    SharedDescendants,
+}
+
+pub type GrimpResult<T> = Result<T, GrimpError>;
+
+impl From<GrimpError> for PyErr {
```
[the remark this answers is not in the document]
Peter554: No strong feeling about this one. Happy to change it.

rust/src/lib.rs
```diff
@@ -1,29 +1,43 @@
...
 
-create_exception!(_rustgrimp, NoSuchContainer, pyo3::exceptions::PyException);
-
 #[pyclass(name = "Graph")]
 struct GraphWrapper {
     _graph: Graph,
 }
 
+impl GraphWrapper {
+    fn get_visible_module_by_name(&self, name: &str) -> Result<&Module, GrimpError> {
```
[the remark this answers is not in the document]
Peter554: I think it was the "visible" part that had me putting this here. Currently the methods within `graph/` return modules whether they are visible or not. The filtering to visible/invisible is kept as a concern for `lib.rs`.

That said, we have the `visible()` method on `ModuleIterator` within `graph/`, so maybe the division isn't as clear as I'd imagined.

Tldr; I'm happy to move it, but no strong feeling.

rust/src/graph/mod.rs
```diff
@@ -0,0 +1,137 @@
...
+use derive_new::new;
+use getset::{CopyGetters, Getters};
+use lazy_static::lazy_static;
+use rustc_hash::{FxHashMap, FxHashSet};
+use slotmap::{new_key_type, SecondaryMap, SlotMap};
+use std::sync::RwLock;
+use string_interner::backend::StringBackend;
+use string_interner::{DefaultSymbol, StringInterner};
+
+pub mod direct_import_queries;
```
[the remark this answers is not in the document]
Peter554: Good spot. I think I was imaging that if the rust crate was published on its own then all the method of `Graph` would be public (`pub`), but the pathfinding module would probably be internal to the crate (`pub(crate)`). That said, I don't think we ever envisage the rust code to be published independently, so probably it could all just be `pub` for consistency.

rust/src/graph/mod.rs
Peter554: No strong feeling, but your suggestion is probably a little nicer 👍

rust/src/graph/mod.rs
Peter554: Yes true your suggestion is much better 👍 Let's change it.

rust/src/graph/mod.rs
Peter554: Yes true your suggestion is better 👍 Let's change it.

rust/src/lib.rs
Peter554: Interesting I forgot about this one. I thought it was necessary since `FxHashSet` doesn't implement `IntoPy`, but presumably your code was working before so I'm now unsure 🤔

rust/src/lib.rs
Peter554: Happy to change this - I think it's a copy over from pyimports.

rust/src/lib.rs
Peter554: The `skip(1)` removes the self, leaving just the ancestors. I'll comment that in the code?

seddonym: Sounds good!

rust/src/lib.rs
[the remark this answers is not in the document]
seddonym: Generally, I think we should keep `GraphWrapper` focused on py03 concerns and try to move everything else into the `Graph` object.

PR #177, opened 2025-01-18
More benchmarks

Peter554:
This PR adds some more benchmarks. I'm making some progress on the rust graph [here](https://github.com/seddonym/grimp/pull/175), but it would be helpful to have a few more benchmarks in place so that we can measure more things. Currently we only have building the graph and illegal layers - there's quite a lot of ground inbetween and the gap feels a bit daunting! 

It also adds a fix for the existing benchmarks, when building the large graph.

### DIFF
--- a/tests/benchmarking/test_benchmarking.py
+++ b/tests/benchmarking/test_benchmarking.py
@@ -6,6 +6,15 @@ from grimp import PackageDependency, Route
 import grimp
 
 
+def _run_benchmark(benchmark, fn, *args, **kwargs):
+    if hasattr(benchmark, "pedantic"):
+        # Running with pytest-benchmark
+        return benchmark.pedantic(fn, args=args, kwargs=kwargs, rounds=3)
+    else:
+        # Running with codspeed.
+        return benchmark(fn, *args, **kwargs)
+
+
 @pytest.fixture(scope="module")
 def large_graph():
     raw_json = (Path(__file__).parent / "large_graph.json").read_text()
@@ -40,13 +49,7 @@ def test_build_django_uncached(benchmark):
 
     In this benchmark, the cache is turned off.
     """
-    fn = lambda: grimp.build_graph("django", cache_dir=None)
-    if hasattr(benchmark, "pedantic"):
-        # Running with pytest-benchmark
-        benchmark.pedantic(fn, rounds=3)
-    else:
-        # Running with codspeed.
-        benchmark(fn)
+    _run_benchmark(benchmark, grimp.build_graph, "django", cache_dir=None)
 
 
 def test_build_django_from_cache(benchmark):
@@ -57,34 +60,23 @@ def test_build_django_from_cache(benchmark):
     """
     # Populate the cache first, before beginning the benchmark.
     grimp.build_graph("django")
-
-    fn = lambda: grimp.build_graph("django")
-    if hasattr(benchmark, "pedantic"):
-        # Running with pytest-benchmark
-        benchmark.pedantic(fn, rounds=3)
-    else:
-        # Running with codspeed.
-        benchmark(fn)
+    _run_benchmark(benchmark, grimp.build_graph, "django")
 
 
 def test_top_level_large_graph(large_graph, benchmark):
-    result = benchmark(
-        lambda: large_graph.find_illegal_dependencies_for_layers(
-            layers=TOP_LEVEL_LAYERS,
-            containers=("mypackage",),
-        )
+    result = _run_benchmark(
+        benchmark,
+        large_graph.find_illegal_dependencies_for_layers,
+        layers=TOP_LEVEL_LAYERS,
+        containers=("mypackage",),
     )
     assert result == set()
 
 
 def test_deep_layers_large_graph(large_graph, benchmark):
-    fn = lambda: large_graph.find_illegal_dependencies_for_layers(layers=DEEP_LAYERS)
-    if hasattr(benchmark, "pedantic"):
-        # Running with pytest-benchmark
-        result = benchmark.pedantic(fn, rounds=3)
-    else:
-        # Running with codspeed.
-        result = benchmark(fn)
+    result = _run_benchmark(
+        benchmark, large_graph.find_illegal_dependencies_for_layers, layers=DEEP_LAYERS
+    )
     assert result == {
         PackageDependency(
             importer=f"{DEEP_PACKAGE}.application.3242334296.2454157946",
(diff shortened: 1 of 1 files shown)


tests/benchmarking/test_benchmarking.py
```diff
@@ -68,6 +68,52 @@ def test_build_django_from_cache(benchmark):
...
+
+
+def test_find_upstream_modules(large_graph, benchmark):
+    result = benchmark(large_graph.find_upstream_modules, DEEP_LAYERS[0], as_package=True)
+    assert len(result) == 2159
+
+
+def test_find_shortest_chain(large_graph, benchmark):
+    result = benchmark(large_graph.find_shortest_chain, DEEP_LAYERS[0], DEEP_LAYERS[1])
+    assert result is not None and len(result) == 5
```
Peter554, line 88: I think we don't need to assert against the exact result here. This file is for benchmarks, not full tests, so a quick sanity check seems fine to me.

tests/benchmarking/test_benchmarking.py
```diff
@@ -75,127 +74,7 @@ def test_top_level_large_graph(large_graph, benchmark):
             containers=("mypackage",),
         )
     )
-    assert result == {
```
Peter554: ♻️ https://github.com/seddonym/grimp/pull/177#discussion_r1921050223

tests/benchmarking/test_benchmarking.py
Peter554: Another option could be to use snapshots, so that we can assert the exact results but they are not cluttering the tests https://github.com/syrupy-project/syrupy

tests/benchmarking/test_benchmarking.py
seddonym: I'd rather we didn't remove these - we don't actually assert on the exact results of a large graph like this anywhere else and it does give me more confidence that things are happening correctly. Happy for it to be moved out of benchmarks if you feel strongly, but maybe we should leave that for a separate PR.

tests/benchmarking/test_benchmarking.py
```diff
@@ -67,6 +67,52 @@ def test_build_django_from_cache(benchmark):
         benchmark(fn)
 
 
+def test_find_descendants(large_graph, benchmark):
```
seddonym, line 70: Might be worth using benchmark.pedantic for each of these, like with the others, so we get to average out the runs. That will only apply to a local run.

tests/benchmarking/test_benchmarking.py
seddonym: It would be helpful to have it somewhere, at some point (to check feature parity between this and the Rust graph implementation) but I take your point, it could go somewhere else. Don't feel strongly about syrupy either way.

tests/benchmarking/test_benchmarking.py
```diff
@@ -67,175 +67,68 @@ def test_build_django_from_cache(benchmark):
...
+def test_find_shortest_chain__no_chain(large_graph, benchmark):
+    result = benchmark(
+        large_graph.find_shortest_chain,
+        DEEP_LAYERS[0],
+        "mypackage.data.vendors.4053192739.6373932949",
+    )
+    assert result is None
+
+
+def test_find_shortest_chains(large_graph, benchmark):
```
seddonym, line 99: Mind grouping these tests for the same method into a `TestFindShortestChains` class?

seddonym (commented):
Thanks for looking at this, great addition.

For some reason the layer checks have become a lot slower though, any idea why? Maybe it's worth separating out commits adding the new benchmarks so we can add ASAP, then keep the other changes separate. Maybe it's a blip.

tests/benchmarking/test_benchmarking.py
Peter554: I tried asserting against the exact result here, but it doesn't seem possible - `find_shortest_chain`/`find_shortest_chains` seem to return different results between runs? I guess there must be multiple chains of equal length?

tests/benchmarking/test_benchmarking.py
Peter554: Okay - put this back

tests/benchmarking/test_benchmarking.py
```diff
@@ -13,6 +13,7 @@ def large_graph():
     graph = ImportGraph()
 
     for importer, importeds in graph_dict.items():
+        graph.add_module(importer)
```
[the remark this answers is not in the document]
Peter554: I'm not entirely sure I understand why this changed the results and caused the performance slowdown.

One thing that can be happening - the graph is now bigger. E.g. consider

```json
{
  "pkg.a": ["pkg.b"],
  "pkg.c": []
}
```

In the above JSON, the `pkg.c` module would not have been added before.

I've reordered the commits, so that the changes caused e.g. in number of modules becomes more obvious now.  

Even though I don't fully understand why this happens entirely, I am entirely confident in this change to the benchmarks file. I.e. we certainly should be calling  `graph.add_module(importer)`, and the import path that we now find is correct (agrees with pyimports). So I think there's no harm to avoid merging this.

tests/benchmarking/test_benchmarking.py
Peter554: Done it.

I'm not fully sure what this does. Even without pedantic pytest-benchmark will run many iterations and take an average. What's the difference between between iterations and rounds here?

tests/benchmarking/test_benchmarking.py
```diff
@@ -217,7 +338,7 @@ def test_deep_layers_large_graph(large_graph, benchmark):
 
 def test_find_descendants(large_graph, benchmark):
     result = _run_benchmark(benchmark, large_graph.find_descendants, "mypackage")
-    assert len(result) == 17348
+    assert len(result) == 28222
```
Peter554, line 341: FYI @seddonym - see here that fixing the graph increases the number of modules. I.e. we were missing many modules before.

tests/benchmarking/test_benchmarking.py
```diff
@@ -241,7 +362,6 @@ def test_chain_found(self, large_graph, benchmark):
         )
         assert result is not None
 
-    @pytest.mark.xfail("grimp.exceptions.ModuleNotPresent")
```
Peter554: FYI @seddonym - see here that fixing the graph means that the module `mypackage.data.vendors.4053192739.6373932949` now exists!

tests/benchmarking/test_benchmarking.py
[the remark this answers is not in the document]
Peter554: Okay I understand it now.

```py
result = large_graph.find_illegal_dependencies_for_layers(
    layers=(
        # mypackage.foo and mypackage.bar do not exist!!!
        "foo",
        "bar",
    ),
    containers=("mypackage",),
)
assert result == set()
```

The above passes ☝️ So it looks like `find_illegal_dependencies_for_layers` does not error if the layer modules cannot be found in the graph. Presumably this is by design?

If we look at large graph JSON we have e.g.

```json
"mypackage.domain": [],
```

This  is the only occurence of `"mypackage.domain"` in that file. So - `mypackage.domain` is not importing anything, and nothing is importing it - hence the module is not added to the graph.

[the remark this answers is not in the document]
Peter554: So sadly the `test_top_level_large_graph` was not doing anything before 😅  Hence why it is now slower - because it actually does some work!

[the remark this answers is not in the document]
seddonym: Ha ok, makes sense. 

The silent failure is by design only to the point of feature parity, but I think it would be worth changing it at some point. Not now though.

seddonym (approved):
Great stuff

PR #178, opened 2025-01-20
Add benchmark for copying graph

Peter554:
In some of the algorithms we rely on copying the graph. Let's benchmark that!


seddonym (approved):
Good idea!

PR #179, opened 2025-01-21
Benchmark layers kept

Peter554:
I think it's helpful to benchmark this case, as it might be possible to take a shortcut when the contract is kept (I want to do this in the rust implementation, should be _much_ faster when the contract is kept). We want to optimise for this case, since in most cases most contracts will be kept (CI is green most of the time).


seddonym (approved):
Nice idea.

PR #180, opened 2025-02-06
Add benchmark for "..." in graph.modules

Peter554:
In the rust backed implementation this is significantly slower (currently), so lets add a benchmark to track this.

Related: https://github.com/seddonym/grimp/pull/175


seddonym (approved):
Great, thanks for adding

PR #182, opened 2025-02-07
Improve performance of graph.modules

Peter554:


src/grimp/application/ports/graph.py
```diff
@@ -26,7 +27,7 @@ class ImportGraph(abc.ABC):
 
     @property
     @abc.abstractmethod
-    def modules(self) -> Set[str]:
+    def modules(self) -> ABCSet[str]:
```
Peter554, line 30: Very strictly speaking I think this is a breaking change on the python side, but really a very minor one...

rust/src/lib.rs
```diff
@@ -48,12 +49,8 @@ impl GraphWrapper {
...
     }
 
-    pub fn get_modules(&self) -> HashSet<String> {
-        self._graph
-            .all_modules()
-            .visible()
-            .names_as_strings()
-            .collect()
+    pub fn get_modules(&self) -> GraphModuleNames {
+        self._graph.all_modules().visible().names().collect()
```
Peter554, line 53: There is still a copy going on here, but now we only have to copy the symbols and not the entire strings. It's quite a lot faster.

src/grimp/application/ports/graph.py
Peter554: > Very strictly speaking I think this is a breaking change on the python side

For example, an `ABCSet` is not an instance of a `TypingSet`:

```py
from typing import Set as TypingSet
from collections.abc import Set as ABCSet

class S(ABCSet):
    def __contains__(self, x):
        return False

    def __len__(self):
        return 0

    def __iter__(self):
        return iter([])

assert isinstance(set(), TypingSet)
# assert isinstance(S(), TypingSet)

```

PR #183, opened 2025-02-07
Ensure fair benchmarks

Peter554:


Peter554:
Closing since we merged https://github.com/seddonym/grimp/pull/184 which covers this

PR #185, opened 2025-02-07
Faster graph clone

Peter554:


tests/benchmarking/test_benchmarking.py
```diff
@@ -20,7 +20,12 @@ def large_graph():
     for importer, importeds in graph_dict.items():
         graph.add_module(importer)
         for imported in importeds:
-            graph.add_import(importer=importer, imported=imported)
+            graph.add_import(
+                importer=importer,
+                imported=imported,
+                line_number=1,
+                line_contents=f"import {imported}",
```
Peter554, line 27: This change introduces a regression in the benchmark for `test_copy_graph`, but this is reasonable since now the graph object is much bigger!

rust/src/graph/graph_manipulation.rs
```diff
@@ -118,7 +118,7 @@ impl Graph {
         &mut self,
         importer: ModuleToken,
         imported: ModuleToken,
-        line_number: usize,
+        line_number: u16,
```
Peter554, line 121: This gives a small but measurable performance benefit. I think people aren't really having import statements past line 65535 😅

seddonym: I agree with the change - but I can believe there are modules out there in the wild that get that big. How would Grimp behave if it did encounter such a module?

Ideally we should handle it gracefully, but as long as we have the new behaviour captured in a ticket, and the change mentioned in the CHANGELOG, I'm happy.

seddonym (changes requested):
Good stuff - I think the change to `line_number` needs at least a mention in the CHANGELOG though, would you mind adding that?

rust/src/graph/graph_manipulation.rs
Peter554: > How would Grimp behave if it did encounter such a module?

Yes maybe it's worth a bit of thought.

It's quite a micro-optimisation anyways, so I think instead I'll just opt to use u32 here instead - I'm near certain that nobody will overflow that!

Peter554: (With u32 I wouldn't update the changelog)

seddonym: > I'm near certain that nobody will overflow that!

Agreed 😄

PR #186, opened 2025-02-08
Add clippy to lint rust code

Peter554:
* https://github.com/rust-lang/rust-clippy
* https://doc.rust-lang.org/clippy/index.html

Clippy is a very popular linter for rust code. The default configuration is good. Clippy can help us to catch issues related to code correctness, style and performance. It's often possible to apply automatic fixes via the `--fix` flag.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -50,7 +50,7 @@ jobs:
         run: cargo test --no-default-features
         working-directory: ./rust
       - name: Run linter (clippy)
-        run: cargo clippy --all-targets --all-features
+        run: cargo clippy --all-targets --all-features -- -D warnings
         working-directory: ./rust
 
   benchmarks:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -9,8 +9,8 @@ on:
   workflow_dispatch:
 
 jobs:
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -9,8 +9,8 @@ on:
   workflow_dispatch:
 
 jobs:
-  tests:
-    name: Python ${{ matrix.python-version }}, ${{ matrix.os }}
+  check_python:
+    name: Check Python ${{ matrix.python-version }}, ${{ matrix.os }}
```
Peter554, line 13: I've gone for the name "check" here rather than "test", since we're also running black, mypy, flake8 etc.

.github/workflows/main.yml
```diff
@@ -50,7 +50,7 @@ jobs:
         run: cargo test --no-default-features
         working-directory: ./rust
       - name: Run linter (clippy)
-        run: cargo clippy --all-targets --all-features
+        run: cargo clippy --all-targets --all-features -- -D warnings
```
Peter554, line 53: There are currently no violations - mainly because I've been using clippy when developing the rust graph on my branch.

seddonym (approved):

PR #187, opened 2025-02-08
Add `as_packages` parameter to `find_shortest_chain`

Peter554:
The implementation of the pathfinding already supports this, so let's add this to the public API!

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -216,11 +216,15 @@ Methods for analysing import chains
     :return: All the modules that are imported (even indirectly) by the supplied module.
     :rtype: A set of strings.
 
-.. py:function:: ImportGraph.find_shortest_chain(importer, imported)
+.. py:function:: ImportGraph.find_shortest_chain(importer, imported, as_packages=False)
 
     :param str importer: The module at the start of a potential chain of imports between ``importer`` and ``imported``
         (i.e. the module that potentially imports ``imported``, even indirectly).
     :param str imported: The module at the end of the potential chain of imports.
+    :param bool as_packages: Whether to treat the supplied modules as individual modules,
+         or as packages (including any descendants, if there are any). If
+         treating them as packages, all descendants of ``importer`` and
+         ``imported`` will be checked too.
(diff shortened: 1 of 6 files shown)


rust/src/lib.rs
```diff
@@ -296,17 +296,18 @@ impl GraphWrapper {
...
         importer: &str,
         imported: &str,
+        as_packages: bool,
     ) -> PyResult<Option<Vec<String>>> {
         let importer = self.get_visible_module_by_name(importer)?.token();
         let imported = self.get_visible_module_by_name(imported)?.token();
         Ok(self
             ._graph
-            .find_shortest_chain(importer, imported, false)?
+            .find_shortest_chain(importer, imported, as_packages)?
```
Peter554, line 310: The implementation already supports it - we just need to pass the parameter through!

src/grimp/application/ports/graph.py
```diff
@@ -220,11 +220,19 @@ def find_upstream_modules(self, module: str, as_package: bool = False) -> Set[st
         raise NotImplementedError
 
     @abc.abstractmethod
-    def find_shortest_chain(self, importer: str, imported: str) -> tuple[str, ...] | None:
+    def find_shortest_chain(
+        self, importer: str, imported: str, as_packages: bool = False
```
Peter554, line 224: Default to false ensures backwards compatibility.

docs/usage.rst
```diff
@@ -221,6 +221,10 @@ Methods for analysing import chains
     :param str importer: The module at the start of a potential chain of imports between ``importer`` and ``imported``
         (i.e. the module that potentially imports ``imported``, even indirectly).
     :param str imported: The module at the end of the potential chain of imports.
+    :param bool as_packages: Whether to treat the supplied modules as individual modules,
```
seddonym, line 224: We should add this to the signature on l. 219 too. (While we're at it, would you mind adding it to the signature of `find_shortest_chains` on l. 231, too? I missed that one.)

seddonym (approved):
👏🏻  Tiny tweak needed to docs but otherwise good to go.

seddonym (approved):

Issue #189, opened 2025-02-16
Find circular dependencies


K4liber:
Adding a functionality for finding circular dependencies between modules/packages.

seddonym:
Thanks for looking at this - just FYI I am on the fence about adding this to Grimp - when I've looked at this before, the results you get on a real code base are not as helpful as you might expect. (You can have potentially millions of results just because if a codebase isn't layered, then there are potentially very many cycles.)

There is always the option of converting the graph to networkx as documented [here](https://grimp.readthedocs.io/en/v2.3/networkx.html), if analysis is needed that Grimp doesn't support.

I would be open to reviewing something, but it may be thorny to figure out how to present useful results, on real code bases.

seddonym:
Oh, I've just realised what this is related to - https://github.com/seddonym/import-linter/pull/250. Sorry if I'm giving you mixed messages 😄 

As I mentioned in [my other comment](https://github.com/seddonym/import-linter/pull/250#discussion_r1929011517), I think we might want to do some cycle-detection in Grimp. But I'm less convinced that this should be part of Grimp's public API.

Instead I would imagine we would have a high level method a bit like `find_illegal_dependencies_for_layers` that returns a meaningful result that Import Linter could be a thin wrapper around. It's probably worth agreeing on what that API might look like - and thoroughly defining what it would actually return, in the case of complex import graphs - before going too far down the implementation route.

K4liber:
@seddonym thanks for the comments. Looking at the current API of Grimp, I can see that a public method `def find_cycles(self) -> list[list[str]]` has a higher generality level than other public methods. Something more specific should fit better to the current code base.

@Peter554 already took a look into the Draft of implementation and he has some good insights. Before any implementation let's agree: 1. If you would like to have such a functionality in Grimp and 2. How the method signature should look like.

When it comes to the point nr 1, I would really encourage you guys to welcome such functionality in the Grimp since it will allows to implement a new contract in the `import-linter` (https://github.com/seddonym/import-linter/pull/250) with the efficient (`Rust`) and independent (using Grimp instead of third-party package) approach. The new contract is a really friendly contract from the user perspective since it does not require any considerations on the layers hierarchy. It does simply not allow for any bidirectional dependency on modules or packages level. If I would start using `import-linter` for my project, such a contract could be the easiest start to improve the dependencies structure.

The proposed method signatures (point 2.):

[...]

PR #190, opened 2025-02-16
Find circular dependencies

K4liber:
Solving #189


rust/src/graph/cycles.rs
```diff
@@ -0,0 +1,138 @@
+/*
+TODO K4liber: description
+
+Finds all cycles using Johnson's algorithm.
```
Peter554, line 4: 🤔 What's the advantage here of Johnson's algorithm over just using the existing pathfinding function? E.g. why not `find_shortest_path_bidirectional(&graph, &start.into(), &start.into(), ...)` - I think that should work? Adding a new algorithm like this adds quite some complexity that needs to be maintained - especially since you algorithm basically transforms the graph to build an entirely new structure (`_DirectedGraph`).

rust/src/graph/cycles.rs
```diff
@@ -0,0 +1,138 @@
...
+                }
+            }
+        }
+    }
+}
+
+impl Graph {
+    pub fn find_cycles(
+        &self
+    ) -> Vec<Vec<String>> {
```
Peter554, line 122: 🤔 I'm not sure we want to work with `String` here . Inside `Graph` I've designed the code to work in terms of `ModuleToken`s. We generally translate back to strings just for the python layer (i.e. in `GraphWrapper` in `lib.rs`).

rust/src/lib.rs
```diff
@@ -420,6 +420,12 @@ impl GraphWrapper {
         self.convert_package_dependencies_to_python(py, illegal_dependencies)
     }
 
+    pub fn find_cycles(
+        &self
+    ) -> PyResult<Vec<Vec<String>>> {
```
Peter554, line 425: 🤔  To find all cycles like this sounds like it could be very computationally expensive for large graphs. Should we instead just give one start module/package to work from? e.g. `find_cycles(&self, module: &str, as_package: bool)` might be a more reasonable API.

Peter554 (commented):
I left a couple of comments, since this looks interesting. (Who am I and why am I commenting :grin: - I’m the one who wrote most of this new rust graph implementation).

I’m a bit unsure if we actually want this in grimp and on the exact public API we’d want - I’d leave that part for @seddonym to consider.

(I see this is still a draft, but I thought some early feedback can still be useful - in particular it seems like we should align a bit on whether we want this and the public API, to avoid wasting your time)

rust/src/graph/cycles.rs
Peter554: > E.g. why not find_shortest_path_bidirectional(&graph, &start_modules, &start_modules, ...) - I think that should work?

To make my point:

```diff
diff --git a/rust/src/graph/pathfinding.rs b/rust/src/graph/pathfinding.rs
index 55e9450..004338e 100644
--- a/rust/src/graph/pathfinding.rs
+++ b/rust/src/graph/pathfinding.rs
@@ -39,6 +39,37 @@ pub fn find_shortest_path_bidirectional(
         return Err(GrimpError::SharedDescendants);
     }
 
+    _find_shortest_path(
+        graph,
+        from_modules,
+        to_modules,
+        excluded_modules,
+        excluded_imports,
+    )
+}
+
+pub fn find_shortest_loop_bidirectional(
+    graph: &Graph,
+    from_modules: &FxHashSet<ModuleToken>,
+    excluded_modules: &FxHashSet<ModuleToken>,
+    excluded_imports: &FxHashMap<ModuleToken, FxHashSet<ModuleToken>>,
+) -> GrimpResult<Option<Vec<ModuleToken>>> {
+    _find_shortest_path(
+        graph,
+        from_modules,
+        from_modules,
+        excluded_modules,
+        excluded_imports,
+    )
+}
+
+fn _find_shortest_path(
+    graph: &Graph,
+    from_modules: &FxHashSet<ModuleToken>,
+    to_modules: &FxHashSet<ModuleToken>,
+    excluded_modules: &FxHashSet<ModuleToken>,
+    excluded_imports: &FxHashMap<ModuleToken, FxHashSet<ModuleToken>>,
+) -> GrimpResult<Option<Vec<ModuleToken>>> {
     let mut predecessors: FxIndexMap<ModuleToken, Option<ModuleToken>> = from_modules
         .clone()
         .into_iter()
@@ -123,3 +154,30 @@ fn import_is_excluded(
             .contains(to_module)
     }
 }
+
+#[cfg(test)]
+mod test_find_shortest_loop_bidirectional {
+    use super::*;
+
+    #[test]
+    fn test_finds_loop() -> GrimpResult<()> {
+        let mut graph = Graph::default();
+        let foo = graph.get_or_add_module("foo").token;
+        let bar = graph.get_or_add_module("bar").token;
+        let baz = graph.get_or_add_module("baz").token;
+        graph.add_import(foo, bar);
+        graph.add_import(bar, baz);
+        graph.add_import(baz, foo);
+
+        let path = find_shortest_loop_bidirectional(
+            &graph,
+            &foo.into(),
+            &FxHashSet::default(),
+            &FxHashMap::default(),
+        )?;
+
+        assert_eq!(path, Some(vec![foo, bar, baz, foo]));
+
+        Ok(())
+    }
+}

```

K4liber:
Thanks @Peter554 for the comments. As you suggest, let's discuss the justification and a signature of the method before we go into more details. I started the conversation here: https://github.com/seddonym/grimp/issues/189#issuecomment-2676079736

PR #191, opened 2025-02-22
Fix CodSpeed action by updating to CodSpeedHQ/action@v3

adriencaccia:
Hey @seddonym, @Peter554, I am a co-founder @CodSpeedHQ.

The latest CodSpeed runs on the repo were not processed because you are using an older version of the action. Upgrading to the v3 will make it work!


seddonym (approved):
Thanks so much!

PR #192, opened 2025-02-25
Add module expressions

seddonym:
Adds module expressions - most of the work is from https://github.com/seddonym/grimp/pull/188.

Still to do:

- Make sure the Python tests are comprehensive (at the moment some of them are purely in Rust).
- Consider whether the Python API for import expressions should use two-tuples or arrow-strings (i.e. `"importer -> imported"`).

### DIFF
--- a/rust/src/module_expressions.rs
+++ b/rust/src/module_expressions.rs
@@ -1,0 +1,192 @@
+use crate::errors::{GrimpError, GrimpResult};
+use itertools::Itertools;
+use lazy_static::lazy_static;
+use regex::Regex;
+use std::fmt::Display;
+use std::str::FromStr;
+
+lazy_static! {
+    static ref MODULE_EXPRESSION_PATTERN: Regex =
+        Regex::new(r"^(\w+|\*{1,2})(\.(\w+|\*{1,2}))*$").unwrap();
+}
+
+/// A module expression is used to refer to sets of modules.
+///
+/// - `*` stands in for a module name, without including subpackages.
+/// - `**` includes subpackages too.
+#[derive(Debug, Clone)]
+pub struct ModuleExpression {
+    expression: String,
+    pattern: Regex,
+}
+
+impl Display for ModuleExpression {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.expression)
+    }
+}
+
+impl AsRef<str> for ModuleExpression {
+    fn as_ref(&self) -> &str {
+        &self.expression
+    }
+}
+
+impl FromStr for ModuleExpression {
+    type Err = GrimpError;
+
+    fn from_str(expression: &str) -> Result<Self, Self::Err> {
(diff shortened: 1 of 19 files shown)


Peter554:
> Consider whether the Python API for import expressions should use two-tuples or arrow-strings (i.e. "importer -> imported").

Incase it gets buried: x-posting my remaining concern about the arrow strings here -> https://github.com/seddonym/grimp/pull/188#discussion_r1969486630. I'm fine with whatever decision you want to take here though.

rust/src/module_expressions.rs
```diff
@@ -0,0 +1,191 @@
...
+            if part == "*" {
+                pattern_parts.push(part.replace("*", r"[^\.]+"))
+            } else if part == "**" {
+                pattern_parts.push(part.replace("**", r"[^\.]+(?:\.[^\.]+)*?"))
+            } else {
+                pattern_parts.push(part.to_owned());
+            }
+        }
+
+        Ok(Regex::new(&(r"^".to_owned() + &pattern_parts.join(r".") + r"$")).unwrap())
```
Peter554, line 77: Should be `Ok(Regex::new(&(r"^".to_owned() + &pattern_parts.join(r"\.") + r"$")).unwrap())` i.e. `\.` instead of `.`. Sorry, this one slipped past me 🙈

src/grimp/adaptors/graph.py
```diff
@@ -102,6 +113,21 @@ def get_import_details(self, *, importer: str, imported: str) -> List[DetailedIm
             imported=imported,
         )
 
+    def find_matching_direct_imports(self, import_expression: str) -> List[Import]:
+        try:
+            importer_expression, imported_expression = import_expression.split(" -> ")
```
Peter554, line 118: 🐼 Should we split and then strip each side? Do we mind if people do `foo->bar` or `foo     ->    bar`?

rust/src/module_expressions.rs
```diff
@@ -0,0 +1,196 @@
...
+
+        Ok(Self {
+            expression: expression.to_owned(),
+            pattern: Self::create_pattern(expression)?,
+        })
+    }
+}
+
+const MODULE_NAME_REGEX: &str = r"[^\.]+";
+const MULTIPLE_MODULE_NAMES_REGEX: &str =
```
Peter554, line 61: 🐼  Maybe `MODULE_NAME_PATTERN` is slightly more correct, since it's not technically a `Regex`.

Peter554 (approved):
Thanks for finishing this off, and for finding that bug!

rust/src/module_expressions.rs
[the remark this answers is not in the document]
seddonym: I like it!

seddonym: Good point, I'll change.

src/grimp/adaptors/graph.py
seddonym: Good point - but I think I'm okay with being prescriptive about a consistent format, for now at least.

PR #193, opened 2025-03-01
Add test demoing that not all chains are found

Peter554:
This PR isn't intended to be merged. It's here to demo a behaviour of `find_illegal_dependencies_for_layers` that I think might be misunderstood.

I think in discussions before we'd said that `find_illegal_dependencies_for_layers` finds all chains. I might be misremembering. Anyways, I just want to put this test case here to demo that not all chains are found.

Update: I just found the existing test case `test_longer_forked_routes_dont_appear`, which is basically testing the same thing 🙈 

Why does this happen? The code here https://github.com/seddonym/grimp/blob/8c1bb5dd23087c4095c57347da46ae3110db439d/rust/src/graph/higher_order_queries.rs#L151-L157 pops all the imports along a chain when a chain is found.

Is there a way to find all the chains? I think not really. I believe finding all paths between two nodes in a graph is an NP hard problem. So probably the existing behavior is the best we can do.

Fine to close this PR if you like. I just wanted to put this here as a reminder of this behaviour and to provoke thought.

I wonder - should this behaviour be documented somewhere? I think the grimp docs don't explicitly say that all chains will be found, but equally they don't call out explicitly that some chains will not be found.

### DIFF
--- a/tests/unit/adaptors/graph/test_layers.py
+++ b/tests/unit/adaptors/graph/test_layers.py
@@ -958,3 +958,43 @@ def _pairwise(iterable):
     a, b = itertools.tee(iterable)
     next(b, None)
     return zip(a, b)
+
+
+def test_demo_not_all_chains_found():
+    graph = ImportGraph()
+
+    # Add routes
+    # ==========
+    #
+    #     a       e
+    #     ↓       ↑
+    #     b → c → d
+    #     ↓       ↑
+    #     x → y → z
+    #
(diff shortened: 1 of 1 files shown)


seddonym:
> I wonder - should this behaviour be documented somewhere? I think the grimp docs don't explicitly say that all chains will be found, but equally they don't call out explicitly that some chains will not be found.

It is documented here, right? https://grimp.readthedocs.io/en/stable/usage.html#return-value

Issue #195, opened 2025-03-11
Improve error handling in ModuleTokenIterator


mariusgoertz:
I have a big repository. I am using grimp for package dependency analytics.
For that i am creating an ImportGraph and squashing it to depth=2 before using the [grimp analytics functions](https://grimp.readthedocs.io/en/stable/usage.html#methods-for-analysing-direct-imports). 

After the rust rewrite in version 3.6, i am getting 

the following error when calling any of the graph analytic functions: 

```
thread '<unnamed>' panicked at src/graph/mod.rs:125:43:
called `Option::unwrap()` on a `None` value
```

Was not able to reproduce this by a minified example yet, so it might be something special casy about my repository structure, but without more context about the failed module in the error traceback, i am a bit lost.

Peter554:
I believe this is related to the order of operations in the squash module method.

* Back in v3.5 we added imports and then removed descendant modules https://github.com/seddonym/grimp/blob/v3.5/src/grimp/adaptors/graph.py#L105-L114
* In v3.6 we removed descendant modules and then added imports https://github.com/seddonym/grimp/blob/v3.6/rust/src/graph/graph_manipulation.rs#L175-L187

For imports within descendants in the v3.6 version we will run into problems, since we end up adding imports to modules that no longer exist. In v3.6
* Suppose descendant1 imports descendant2
* We will remove descendant1 and descendant2 from the graph
* We will then add the import from descendant1 to descendant2 back into the graph
* The imports map is now corrupted - it contains an import for modules that no longer exist in the graph

I _think_ the fix should be easy, just reverse the order of operations to be consistent with v3.5

```diff
diff --git a/rust/src/graph/graph_manipulation.rs b/rust/src/graph/graph_manipulation.rs
index caed21c..a484fc1 100644
--- a/rust/src/graph/graph_manipulation.rs
+++ b/rust/src/graph/graph_manipulation.rs
@@ -178,11 +178,6 @@ impl Graph {
             })
             .collect();
 
-        // Remove any descendants.
-        for descendant in descendants {
-            self.remove_module(descendant);
-        }
-
         // Add descendants and imports to parent module.
         for imported in modules_imported_by_descendants {
             self.add_import(module, imported);
@@ -192,6 +187,11 @@ impl Graph {
             self.add_import(importer, module);
         }
 
+        // Remove any descendants.
+        for descendant in descendants {
+            self.remove_module(descendant);
+        }
+        
         self.mark_module_squashed(module);
     }
 }
[...]

PR #196, opened 2025-03-11
Fix handling of within-descendant-imports when squashing modules

Peter554:
Fixes https://github.com/seddonym/grimp/issues/195 (I'm pretty sure - hard to be 100% certain without seeing the respository in question).

### DIFF
--- a/tests/unit/adaptors/graph/test_manipulation.py
+++ b/tests/unit/adaptors/graph/test_manipulation.py
@@ -447,3 +447,23 @@ class TestSquashModule:
 
         with pytest.raises(ModuleNotPresent):
             graph.squash_module("foo")
+
+    @pytest.mark.xfail(strict=True, reason="raises PanicException")
+    def test_correctly_handles_imports_within_descendants(self):
+        graph = ImportGraph()
+
+        graph.add_module("animals")
+        graph.add_module("food")
+        graph.add_import(importer="animals.dog", imported="food.chicken")
+        graph.add_import(importer="app.cli", imported="animals.dog")
+        # We want to check that this import within the descendants of `animals` does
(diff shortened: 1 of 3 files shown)


tests/unit/adaptors/graph/test_manipulation.py
```diff
@@ -447,3 +447,23 @@ def test_raises_module_not_present_if_no_module(self):
...
+
+        graph.add_module("animals")
+        graph.add_module("food")
+        graph.add_import(importer="animals.dog", imported="food.chicken")
+        graph.add_import(importer="app.cli", imported="animals.dog")
+        # We want to check that this import within the descendants of `animals` does
+        # not cause problems. If this import is not properly removed then the imports map can
+        # become corrupted, since it will contain an import to modules that no longer exist.
+        # See https://github.com/seddonym/grimp/issues/195 for more details.
+        graph.add_import(importer="animals.dog", imported="animals.base")
```
seddonym, line 462: Should we add imports between deeper descendants too, such as grandchildren?

Peter554:
FYI @seddonym @mariusgoertz

seddonym (approved):
Thanks for fixing so quickly!

tests/unit/adaptors/graph/test_manipulation.py
Peter554: We could. I'm not sure if it would add anything though, I think it shouldn't make any difference. So maybe it just confuses the test case to add more?

mariusgoertz:
> Fixes #195 (I'm pretty sure - hard to be 100% certain without seeing the respository in question).

i checked out the fix and tested it and can confirm its working again now! thanks!

PR #198, opened 2025-04-04
Parallel import scanning (python)

Peter554:


src/grimp/exceptions.py
```diff
@@ -62,6 +62,9 @@ def __eq__(self, other):
             other.text,
         )
 
+    def __reduce__(self):
+        return SourceSyntaxError, (self.filename, self.lineno, self.text)
```
Peter554, line 66: See https://stackoverflow.com/a/36342588

Peter554:
Screenshot 2025-04-04 at 11 42 33

@seddonym This doesn't seem believable ☝️  - I think it can have something to do with https://docs.codspeed.io/instruments/cpu/#system-calls

Benchmarks on my machine:

```
------- benchmark 'test_build_django_from_cache': 2 tests -------
Name (time in ms)                                  Mean
-----------------------------------------------------------------
test_build_django_from_cache (master)     36.0710 (1.0)
test_build_django_from_cache (NOW)              72.1804 (2.00)
-----------------------------------------------------------------

-------- benchmark 'test_build_django_uncached': 2 tests ---------
Name (time in ms)                                   Mean
------------------------------------------------------------------
test_build_django_uncached (NOW)                251.8374 (1.0)
test_build_django_uncached (master)     1,064.6040 (4.23)
------------------------------------------------------------------
```

src/grimp/exceptions.py
seddonym: Probably worth adding a comment to explain why this is necessary.

src/grimp/application/usecases.py
```diff
@@ -122,18 +121,31 @@ def _scan_packages(
         include_external_packages=include_external_packages,
     )
 
-    for found_package in found_packages:
-        for module_file in found_package.module_files:
-            module = module_file.module
+    modules_files_to_scan = {
```
seddonym, line 131: Should this be `module_files_to_scan`?

src/grimp/application/usecases.py
```diff
@@ -184,3 +203,21 @@ def _is_external(module: Module, found_packages: Set[FoundPackage]) -> bool:
         module.is_descendant_of(package_module) or module == package_module
         for package_module in package_modules
     )
+
+
+def _create_chunks(module_files: List[ModuleFile], *, n_chunks: int) -> List[List[ModuleFile]]:
```
seddonym, line 208: 🐼 `module_files` could be `Collection[ModuleFile]` for more flexibility.

seddonym, line 208: 🐼 Always sceptical when I see a list 😉 .

Maybe `tuple[tuple[ModuleFile], ...]` would be a slightly better return type as it's immutable?

Also considered `frozenset[frozenset[ModuleFile]]` but I wonder if hashing might incur a penalty here.

src/grimp/application/usecases.py
```diff
@@ -184,3 +203,21 @@ def _is_external(module: Module, found_packages: Set[FoundPackage]) -> bool:
...
+    chunk_size = math.ceil(len(module_files) / n_chunks)
+    return [module_files[i * chunk_size : (i + 1) * chunk_size] for i in range(n_chunks)]
+
+
+def _scan_imports(
+    import_scanner: AbstractImportScanner,
+    exclude_type_checking_imports: bool,
+    module_files: List[ModuleFile],
```
seddonym, line 216: Could be `Iterable[ModuleFile]` for greater flexibility.

src/grimp/application/usecases.py
```diff
@@ -137,11 +144,23 @@ def _scan_packages(
             else:
                 imports_by_module_file[module_file] = direct_imports
 
-    # TODO Parallelise this part.
-    for module_file in modules_files_to_scan.difference(imports_by_module_file):
-        imports_by_module_file[module_file] = import_scanner.scan_for_imports(
-            module_file.module, exclude_type_checking_imports=exclude_type_checking_imports
+    remaining_modules_files_to_scan = list(
```
seddonym, line 147: I think the `_scan_packages` function has tipped over into becoming too long now - would you mind breaking it up a bit?

src/grimp/application/usecases.py
```diff
@@ -137,11 +144,23 @@ def _scan_packages(
...
+        modules_files_to_scan.difference(imports_by_module_file)
+    )
+    chunked_remaining_modules_files_to_scan = _create_chunks(
+        remaining_modules_files_to_scan, n_chunks=N_CPUS
+    )
+    with parallel_config(n_jobs=N_CPUS):
```
seddonym, line 153: When the cache is full, `remaining_modules_files_to_scan` is empty - if that's the case it's a waste of time to spin up these processes (and it noticeably degrades performance). At the very least we should wrap this in `if remaining_modules_files_to_scan`. I tried that locally and it recovered the degraded performance for a fully populated cache.

Also, `n_jobs` should be `min(N_CPUS, len(remaining_module_files_to_scan))`, right? I tried that locally and it improves the situation if I've only changed one file. It would be good to add a benchmark for that scenario as I don't think we currently measure it.

I'm not sure what the threshold is multiprocessing becomes worthwhile, but we could possibly improve on the logic a little by having a non-parallel track for this too, for small numbers of changes. That said, I tried the approach of spinning up only one extra process for one changed module and it wasn't noticeably slower, so it's possibly not worth it.

seddonym (changes requested):
This is fantastic, thank you! A real game changer.

I would like to address the degraded performance when the cache is warm - but it looks like that is really easily done. I'll open a separate PR to benchmark using the cache for small numbers of changes, and we can rebase off this to check there isn't degraded performance there too.

I'm not yet sold on introducing joblib - would prefer not to introduce third party dependencies unless there's a really strong reason. Could we try this using the standard library instead?

seddonym:
Oh - just remembered one thing, would you mind adding a line to the CHANGELOG too?

Peter554 (commented):
@seddonym I've added a pure multiprocessing version as a fixup now. It does seem to be a bit slower. 

* When running tests locally I notice the difference in the overall time of the test suite.
* When running benchmarks I see that it's quite a bit slower than joblib.

```
------- benchmark 'test_build_django_from_cache':  --------------
Name (time in ms)                                         Mean
-----------------------------------------------------------------
test_build_django_from_cache (joblib)                     37.9565
test_build_django_from_cache (master)                     38.1530
test_build_django_from_cache (multiprocessing)            39.2882
-----------------------------------------------------------------

-------- benchmark 'test_build_django_uncached':  ----------------
Name (time in ms)                                         Mean
------------------------------------------------------------------
test_build_django_uncached (joblib)                       250.2343
test_build_django_uncached (multiprocessing)              421.2143
test_build_django_uncached (master)                     1,066.4864
------------------------------------------------------------------
```

What do you think?

I think it's likely okay to use joblib here.

* joblib has 4000 stars on GitHub, so isn't going anywhere too soon.
* joblib has no dependences beyond python and emphasises robustness over features (see "design choices" https://joblib.readthedocs.io/en/stable/why.html#design-choices)
* Absolute worst case if this dependency breaks we can just remove it again and things get a bit slower.

src/grimp/application/usecases.py
```diff
@@ -186,3 +182,38 @@ def _is_external(module: Module, found_packages: Set[FoundPackage]) -> bool:
         module.is_descendant_of(package_module) or module == package_module
         for package_module in package_modules
     )
+
+
+def _read_imports_from_cache(
```
seddonym, line 190: Thanks for breaking this up, much easier to read.

src/grimp/application/usecases.py
```diff
@@ -14,6 +15,8 @@
 from ..domain.valueobjects import DirectImport, Module
 from .config import settings
 
+N_CPUS = multiprocessing.cpu_count()
```
seddonym, line 18: 🤔 I wonder what the overhead of this is, probably tiny but it might make sense to calculate it at runtime, only if we need it.

Maybe worth tweaking next time we're in this file.

seddonym (approved):
Great stuff, thanks so much for this.

As discussed elsewhere, we should probably add an optimization for small numbers of modules where it's not worth multiprocessing. That will speed up the functional tests, if nothing else.

PR #199, opened 2025-04-04
Benchmark using the cache for a few misses

seddonym:
It's useful to distinguish this scenario from the others, as it's a common situation which we should make sure we optimize for.

### DIFF
--- a/tests/benchmarking/test_benchmarking.py
+++ b/tests/benchmarking/test_benchmarking.py
@@ -1,5 +1,6 @@
 import pytest
 import json
+import importlib
 from pathlib import Path
 from grimp.adaptors.graph import ImportGraph
 from grimp import PackageDependency, Route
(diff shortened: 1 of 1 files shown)


tests/benchmarking/test_benchmarking.py
```diff
@@ -299,14 +300,39 @@ def test_build_django_uncached(benchmark):
...
+    """
+    # Populate the cache first, before beginning the benchmark.
+    grimp.build_graph("django")
+    # Add a module which won't be in the cache.
+    django_path = Path(importlib.util.find_spec("django").origin).parent
+    for i in range(number_of_misses):
+        new_module = django_path / f"new_module_{i}.py"
+        new_module.write_text("from django.db import models")
+
     _run_benchmark(benchmark, grimp.build_graph, "django")
```
Peter554, line 336: I think this doesn't work quite as you're thinking:

* `_run_benchmark` will run the function (2nd argument, here `grimp.build_graph`) over many iterations. 
* The `new_module_*.py` modules will only be new on the first iteration. On later iterations the modules will already be in the cache.
* You can see this in the benchmark results: `test_build_django_from_cache_no_misses`, `test_build_django_from_cache_a_few_misses[2]` and `test_build_django_from_cache_a_few_misses[15]` all have the same time. 

Another more minot thing: These new modules are just a single line - so there will be very little work to do when parsing the AST. I.e. this won't be very realistic.

PR #200, opened 2025-04-04
Address issue with setting up the partial cache in benchmark

seddonym:
Prior to this, we would set up the state for the first iteration, but future iterations would be able to use the fully-populated cache.

### DIFF
--- a/tests/benchmarking/adaptors.py
+++ b/tests/benchmarking/adaptors.py
@@ -1,0 +1,20 @@
+from grimp.adaptors.caching import Cache, CacheMiss
+from typing import Set
+
+from grimp.application.ports.modulefinder import ModuleFile
+from grimp.domain.valueobjects import DirectImport
+
+
+class PrefixMissingCache(Cache):
+    """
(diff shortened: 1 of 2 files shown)


Peter554 (approved):
A bit awkward, but seems a good idea.

I'm a bit confused that the benchmark seems to take the same time as fully from the cache - I would have expected it to be a bit slower. I suppose 2 and 15 are not very big numbers, maybe it makes sense to add a case where we are missing 100, or 1000 cache entries.

seddonym:
> I'm a bit confused that the benchmark seems to take the same time as fully from the cache - I would have expected it to be a bit slower. I suppose 2 and 15 are not very big numbers, maybe it makes sense to add a case where we are missing 100, or 1000 cache entries.

I've added one for 350 misses (about half of Django). It does slow the benchmarking job down quite a bit unfortunately, maybe we can remove it later.

I've also adjusted it so it copies a real module from Django instead of a one-liner, so the parser will have more work to do.

PR #204, opened 2025-04-07
Parallelize codspeed

seddonym:
Parallelizes the codspeed benchmarks.

Benchmarking in CI was taking about 12 minutes, parallelizing it speeds up the time to completion.


seddonym:
Not doing it at the moment, as we should find a way of parallelizing across the build graph ones which take the longest. It looks like pytest test groups doesn't work well with parametrized tests.

PR #209, opened 2025-04-14
Use joblib for more robust parallel import scanning

Peter554:
https://joblib.readthedocs.io/en/stable/parallel.html

Joblib takes some things for us. Relevant here:

* Robust calculation for number of available CPUs.
* Sequential calculation when n_jobs = 1.

And likely other minor things I don't even understand (joblib has put a lot of thought into multi-processing, so we don't have to).


src/grimp/application/usecases.py
```diff
@@ -21,7 +22,7 @@ class NotSupplied:
 
 
 # This is an arbitrary number, but setting it too low slows down our functional tests considerably.
-MIN_NUMBER_OF_MODULES_TO_SCAN_USING_MULTIPROCESSING = 50
+MIN_NUMBER_OF_MODULES_TO_SCAN_USING_MULTIPLE_PROCESSES = 64
```
Peter554, line 25: 🐼🐼🐼 My OCD => 64 is a power of two, just feels nicer

src/grimp/application/usecases.py
```diff
@@ -21,7 +22,7 @@ class NotSupplied:
 
 
 # This is an arbitrary number, but setting it too low slows down our functional tests considerably.
-MIN_NUMBER_OF_MODULES_TO_SCAN_USING_MULTIPROCESSING = 50
```
Peter554, line 24: `MULTIPLE_PROCESSES` instead of `MULTIPROCESSING` to avoid confusion with `multiprocessing` module

src/grimp/application/usecases.py
```diff
@@ -228,19 +229,19 @@ def _create_chunks(module_files: Collection[ModuleFile]) -> tuple[tuple[ModuleFi
     module_files_tuple = tuple(module_files)
 
     number_of_module_files = len(module_files_tuple)
-    n_chunks = _decide_number_of_of_processes(number_of_module_files)
```
Peter554, line 231: Typo (should really have put that in a separate commit...)

src/grimp/application/usecases.py
```diff
@@ -228,19 +229,19 @@ def _create_chunks(module_files: Collection[ModuleFile]) -> tuple[tuple[ModuleFi
...
 
-def _decide_number_of_of_processes(number_of_module_files: int) -> int:
-    if number_of_module_files < MIN_NUMBER_OF_MODULES_TO_SCAN_USING_MULTIPROCESSING:
-        # Don't incur the overhead of multiprocessing.
+def _decide_number_of_processes(number_of_module_files: int) -> int:
+    if number_of_module_files < MIN_NUMBER_OF_MODULES_TO_SCAN_USING_MULTIPLE_PROCESSES:
+        # Don't incur the overhead of multiple processes.
         return 1
-    return min(multiprocessing.cpu_count(), number_of_module_files)
+    return min(joblib.cpu_count(), number_of_module_files)
```
Peter554, line 244: `joblib.cpu_count()` is a thin wrapper around `loky.cpu_count()` https://github.com/joblib/loky/blob/d8bb877b94214883c2b69cf85ae375ea53d6cd17/loky/backend/context.py#L78

src/grimp/application/usecases.py
```diff
@@ -257,20 +258,15 @@ def _scan_chunks(
     )
 
     number_of_processes = len(chunks)
-    if number_of_processes == 1:
-        # No need to spawn a process if there's only one chunk.
-        [chunk] = chunks
-        return _scan_chunk(import_scanner, exclude_type_checking_imports, chunk)
```
Peter554, line 263: No need for this special `number_of_processes == 1` case anymore - joblib will do this automatically. https://joblib.readthedocs.io/en/stable/generated/joblib.Parallel.html#joblib.Parallel

> If 1 is given, no parallel computing code is used at all, and the behavior amounts to a simple python for loop.

seddonym (commented):
Thanks for this - I've merged into a branch in my repo and will make a couple of tweaks before merging to master.

PR #212, opened 2025-05-01
Import Scanning: Split raw import extraction from import resolution

Peter554:
This PR contains a pre-factor to integrating https://github.com/Peter554/pyimportparse into grimp.

This PR separates raw imported object parsing from the import resolution. This will then allow us to replace the raw imported object parsing part with pyimportparse (we can move the import resolution part over to rust in later work).

### DIFF
--- a/src/grimp/adaptors/importscanner.py
+++ b/src/grimp/adaptors/importscanner.py
@@ -1,5 +1,9 @@
+from __future__ import annotations
+
 import ast
+import re
 import logging
+from dataclasses import dataclass
 from typing import Dict, List, Optional, Set, Union
 from ast import NodeVisitor, Import, ImportFrom, If, Attribute, Name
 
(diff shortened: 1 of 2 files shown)


src/grimp/adaptors/importscanner.py
```diff
@@ -32,47 +46,61 @@ def scan_for_imports(
...
 
         try:
-            ast_tree = ast.parse(module_contents)
+            imported_objects = self._get_raw_imported_objects(module_contents)
         except SyntaxError as e:
             raise exceptions.SourceSyntaxError(
                 filename=module_filename,
                 lineno=e.lineno,
                 text=e.text,
             )
```
Peter554, line 58: @seddonym  This is the part that https://github.com/Peter554/pyimportparse could replace.

seddonym (approved):
Thanks for this!

PR #213, opened 2025-05-02
Replace ast with rust

seddonym:
Replaces the use of Python's built in `ast` module with a custom Rust alternative.

It is faster than `ast` primarily because it is not a fully implemented Python parser, but just pulls out the imports.
 
It's based on https://crates.io/crates/pyimportparse, but there were some aspects of it that were inconsistent with how Grimp currently behaves. So for the time being I've copied it in and adapted it.

It's likely that Astral will publish their Python parser as a crate - once that happens we may be able to move to that instead.

### DIFF
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -22,6 +22,10 @@ rustc-hash = "2.1.0"
 indexmap = "2.7.1"
 regex = "1.11.1"
 const_format = "0.2.34"
+pyimportparse = "0.2.1"
+nom = "8.0.0"
+nom_locate = "5.0.0"
+nom-regex = "0.2.0"
 
 [dependencies.pyo3]
 version = "0.24.1"
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -26,6 +26,7 @@ pyimportparse = "0.2.1"
 nom = "8.0.0"
 nom_locate = "5.0.0"
 nom-regex = "0.2.0"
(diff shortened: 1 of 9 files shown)


tests/unit/adaptors/test_importscanner.py
```diff
@@ -927,6 +927,67 @@ def test_get_raw_imports(self):
             ),
         }
 
+    @pytest.mark.xfail(strict=True)
+    def test_multiline_comment_a(self):
```
Peter554, line 931: https://github.com/Peter554/pyimportparse/commit/e7d3d0da843ce3ea36d5a8342d3cb1bbdb32ea05 might be an option

rust/Cargo.toml
```diff
@@ -22,6 +22,11 @@ rustc-hash = "2.1.0"
 indexmap = "2.7.1"
 regex = "1.11.1"
 const_format = "0.2.34"
+pyimportparse = "0.2.1"
+nom = "8.0.0"
+nom_locate = "5.0.0"
+nom-regex = "0.2.0"
```
Peter554, line 28: nom-regex is not being used I think, so we could remove the dep?

tests/unit/adaptors/test_importscanner.py
Peter554: On reflection, I think this might be a better approach -> https://github.com/seddonym/grimp/pull/214

PR #214, opened 2025-05-03
Use ruff to parse imports

Peter554:
Ruff can be installed via git!

You may need to use `CARGO_NET_GIT_FETCH_WITH_CLI=true`

Inspired by https://github.com/RustPython/RustPython/pull/5494

Performance results on my machine (Kraken):

* AST
  * 1 process: 41s
  * Many processes: 11s
* Ruff
  * 1 process: 15s
  * Many processes: 6s

### DIFF
--- a/rust/src/import_parsing.rs
+++ b/rust/src/import_parsing.rs
@@ -1,5 +1,5 @@
-use crate::errors::GrimpResult;
-use ruff_python_ast::statement_visitor::{walk_body, walk_stmt, StatementVisitor};
+use crate::errors::{GrimpError, GrimpResult};
+use ruff_python_ast::statement_visitor::{StatementVisitor, walk_body, walk_stmt};
 use ruff_python_ast::{Expr, Stmt};
 use ruff_python_parser::parse_module;
 use ruff_source_file::{LineIndex, SourceCode};
@@ -73,10 +73,21 @@ pub fn parse_imports_from_code(code: &str) -> GrimpResult<Vec<ImportedObject>> {
 fn parse_imports_from_code_without_line_contents(
     code: &str,
 ) -> GrimpResult<Vec<ImportedObjectWithoutLineContents>> {
-    let ast = parse_module(code).expect("failed to parse python code");
-
     let line_index = LineIndex::from_source_text(code);
     let source_code = SourceCode::new(code, &line_index);
+
+    let ast = match parse_module(code) {
+        Ok(ast) => ast,
+        Err(e) => {
+            let line_number = source_code.line_index(e.location.start()).get();
+            let text = source_code.slice(e.location);
+            Err(GrimpError::ParseError {
+                line_number,
+                text: text.to_owned(),
+                parse_error: e,
+            })?
+        }
+    };
 
     let mut visitor = Visitor::new(source_code);
     walk_body(&mut visitor, &ast.syntax().body);
--- a/rust/src/import_parsing.rs
+++ b/rust/src/import_parsing.rs
@@ -1,0 +1,595 @@
+use crate::errors::GrimpResult;
+use ruff_python_ast::statement_visitor::{walk_body, walk_stmt, StatementVisitor};
+use ruff_python_ast::{Expr, Stmt};
+use ruff_python_parser::parse_module;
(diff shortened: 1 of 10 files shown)


src/grimp/adaptors/importscanner.py
```diff
@@ -141,17 +139,7 @@ def _module_is_package(self, module_filename: str) -> bool:
     @staticmethod
     def _get_raw_imported_objects(module_contents: str) -> Set[_ImportedObject]:
         imported_object_dicts = rust.parse_imported_objects_from_code(module_contents)
-        imported_objects = {_ImportedObject(**d) for d in imported_object_dicts}
-
-        # TODO - remove these lines once we're confident the rust way is consistent with ast.
```
Peter554: I didn't find any discrepancies when checking Kraken, so removing this.

src/grimp/adaptors/importscanner.py
```diff
@@ -50,7 +50,9 @@ def scan_for_imports(
 
         try:
             imported_objects = self._get_raw_imported_objects(module_contents)
-        except SyntaxError as e:
+        except rust.ParseError as e:
+            # TODO What to do with the rust.ParseError?
+            # TODO The line number and text are part of str(e), but is there a nice way to get them?
```
Peter554, line 55: @seddonym I could do with some help to decide what to do here. The problem is that the line number and text are only available as part of `str(e)`, and not as structured fields.

Peter554: Ah I think I found a reasonable solution now, using `pyclass(extends=PyException)`

tests/functional/test_error_handling.py
```diff
@@ -15,9 +15,7 @@ def test_syntax_error_includes_module():
     with pytest.raises(exceptions.SourceSyntaxError) as excinfo:
         build_graph("syntaxerrorpackage", cache_dir=None)
 
-    expected_exception = exceptions.SourceSyntaxError(
-        filename=filename, lineno=5, text="fromb . import two\n"
-    )
+    expected_exception = exceptions.SourceSyntaxError(filename=filename, lineno=5, text="import")
```
Peter554, line 18: The `text` has changed here - the ruff parser reports a different text range for the syntax error than the python ast parser. See `let text = source_code.slice(e.location);` in the `import_parsing.rs` file.

rust/src/import_parsing.rs
```diff
@@ -0,0 +1,595 @@
...
+                i.typechecking_only,
+            )
+        })
+        .collect())
+}
+
+fn parse_imports_from_code_without_line_contents(
+    code: &str,
+) -> GrimpResult<Vec<ImportedObjectWithoutLineContents>> {
+    let ast = parse_module(code).expect("failed to parse python code");
```
seddonym: Do you know if the Python syntax it checks is the same as the version of Python it will be running under?

rust/src/import_parsing.rs
```diff
@@ -0,0 +1,595 @@
...
+use ruff_source_file::{LineIndex, SourceCode};
+use std::fs;
+use std::path::Path;
+
+#[derive(Debug, PartialEq, Eq, Clone)]
+pub struct ImportedObject {
+    pub name: String,
+    pub line_number: usize,
+    pub line_contents: String,
+    pub typechecking_only: bool,
```
seddonym, line 14: One optimization we could consider for the future: we don't actually need to keep track of whether an import is within a type checking guard here (it's deliberately not exposed in Grimp's public API). Instead, we could just take `exclude_type_checking_imports` as an argument and use a (I'm guessing) slightly faster algorithm if it's false.

seddonym (approved):
Awesome stuff - nice work.

I suspect that writing our own parser to parse a subset of Python might ultimately be faster, but I think it's a good practical decision to do it like this at least for the time being. I'm not quite clear on how it interacts with different Python versions though.

rust/src/import_parsing.rs
Peter554: Good question. I'm unsure, but I'd suspect the latest python syntax is parsed. Assuming python doesn't release a new major version, then the syntax should always be backwards compatible? So it should be fine I think.

Issue #219, opened 2025-05-27
3.14 wheels


kasium:
Hi,

do you have already plans to publish 3.14 wheels?

Thanks :-)

PR #220, opened 2025-06-05
Adapt release process to publish 3.14 wheels

seddonym:
As requested https://github.com/seddonym/grimp/issues/219


seddonym:
Thanks for the nudge, yes it's on my list!

PR #224, opened 2025-06-16
Add snapshot test for import scanning on real packages

Peter554:
Cherry-picked from https://github.com/seddonym/grimp/pull/222

This snapshot test captures the results of import scanning against real packages. This gives us a lot of confidence when refactoring the import scanning code (helpful as we move more of this code over to rust).

The snapshot test uses the popular syrupy[^1] pytest plugin.

[^1]: https://github.com/syrupy-project/syrupy

### DIFF
--- a/tests/functional/__snapshots__/test_build_graph_on_real_packages.ambr
+++ b/tests/functional/__snapshots__/test_build_graph_on_real_packages.ambr
@@ -1,0 +1,21457 @@
+# serializer version: 1
+# name: test_build_graph_on_real_package[django]
+  list([
+    dict({
+      'imported': 'django.apps',
+      'importer': 'django',
+    }),
+    dict({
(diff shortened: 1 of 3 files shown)


.gitattributes
```diff
@@ -0,0 +1,3 @@
+# .ambr files should be excluded from repo stats.
+# https://github.com/github-linguist/linguist/blob/main/docs/overrides.md
+*.ambr linguist-vendored
```
Peter554, line 3: 🤔  I'm not 100% sure I've done this correctly. I'm trying to avoid giving myself a 20_000 line contribution for this generated file. 

I chose `linguist-vendored` over `linguist-generated` since `linguist-generated` is "Excluded from stats, hidden in diffs", and we don't want to hide these files from diffs.

PR #227, opened 2025-06-18
Closed layers

Peter554:
(AI assisted PR description)

# Implement Closed Layers Feature

## Overview

This PR introduces the concept of "closed layers" to Grimp, providing an additional constraint mechanism for controlling imports between modules in a Python codebase. 

### Key Behaviour

When a layer is marked as closed, modules in higher layers cannot import modules that exist below the closed layer. This creates a strict boundary that forces higher layers to interact only with the closed layer itself, not with anything beneath it. This is particularly useful for enforcing clean architectural boundaries where lower-level implementation details should be hidden from higher-level components.

## Example Use Case

Consider this layered architecture:

```
app/
  high_level/     # Higher layer
  mid_level/      # Middle layer (closed)
  low_level/      # Lower layer
```

With `mid_level` configured as a closed layer:
- `high_level` can import from `mid_level`
- `high_level` cannot import from `low_level` (must go through `mid_level`)
- `mid_level` can import from `low_level`

This enforces that `high_level` modules can only interact with `low_level` functionality through the `mid_level` interface, preventing dependency leakage and maintaining clear architectural boundaries.

### DIFF
--- a/tests/unit/adaptors/graph/test_layers.py
+++ b/tests/unit/adaptors/graph/test_layers.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import itertools
 import logging
 import re
+from collections.abc import Iterable
 
 import pytest  # type: ignore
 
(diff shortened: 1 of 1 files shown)


seddonym (changes requested):
Great to see this taking shape!

I haven't reviewed the implementation in detail yet, but I like what I see so far. I do think we should adjust it so that the field on the `Layer` is closed (default False), not open (default True). This is because I want closed layers to feel like more of a special case.
 
Happy to discuss if you feel strongly the other way.

src/grimp/domain/valueobjects.py
```diff
@@ -65,14 +65,19 @@ class Layer:
...
+    If layer.open is True then the layer is considered open.
+    This is the default. Else the layer is closed.
     """
 
     module_tails: Set[str]
     independent: bool
+    open: bool
 
-    def __init__(self, *module_tails: str, independent: bool = True) -> None:
+    def __init__(self, *module_tails: str, independent: bool = True, open: bool = True) -> None:
```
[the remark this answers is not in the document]
Peter554: Yes makes sense - updated.

src/grimp/domain/valueobjects.py
```diff
@@ -69,12 +69,14 @@ class Layer:
...
-    def __init__(self, *module_tails: str, independent: bool = True) -> None:
+    def __init__(self, *module_tails: str, independent: bool = True, closed: bool = False) -> None:
         # `object.__setattr__` is needed since the dataclass is frozen.
         object.__setattr__(self, "module_tails", set(module_tails))
         object.__setattr__(self, "independent", independent)
+        object.__setattr__(self, "closed", closed)
 
     def __str__(self) -> str:
-        return f"{self.module_tails}, independent={self.independent}"
+        return f"{self.module_tails}, independent={self.independent}, closed={self.closed}"
```
seddonym, line 82: Could we have test coverage for this?

CHANGELOG.rst
```diff
@@ -2,6 +2,14 @@
 Changelog
 =========
 
+Unreleased
+----------
+
+* Implement closed layers - a feature that prevents imports from higher layers
```
seddonym, line 8: Could we change this to `Add closed layers to layer contract.` - and skip the rest?

rust/src/graph/higher_order_queries.rs
```diff
@@ -208,3 +208,80 @@ impl Graph {
...
+        independent_layer.insert(module_a);
+        independent_layer.insert(module_b);
+
+        let independent_level = Level::new(independent_layer, true);
+
+        let levels = vec![independent_level];
+
+        let permutations = graph.generate_module_permutations(&levels);
+
+        // Modules in independent layer should not import each other
```
seddonym, line 284: Out of interest, why not just do a single assertion on what the result is?

rust/src/graph/higher_order_queries.rs
```diff
@@ -105,6 +105,17 @@ impl Graph {
                         permutations.push((*module, *sibling_module));
                     }
                 }
+
+                // Should not be imported by higher layers if there is a closed layer inbetween.
+                let mut closed = false;
```
seddonym, line 110: Elegant!

rust/src/graph/higher_order_queries.rs
```diff
@@ -287,4 +298,57 @@ mod tests {
...
+
+        let top_level = Level::new(top_layer, false, false);
+        let middle_level = Level::new(middle_layer, false, true); // Closed layer
+        let bottom_level = Level::new(bottom_layer, false, false);
+
+        let levels = vec![top_level, middle_level, bottom_level];
+
+        let permutations = graph.generate_module_permutations(&levels);
+
+        assert_eq!(
```
seddonym, line 328: Personally I think this would be clearer as a single assertion on the result.

rust/src/graph/higher_order_queries.rs
```diff
@@ -102,6 +105,17 @@ impl Graph {
                         permutations.push((*module, *sibling_module));
                     }
                 }
+
+                // Should not be imported by higher layers if there is a closed layer inbetween.
```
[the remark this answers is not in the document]
seddonym: Coming back to it, I agree with how you've done it here. Can you see any reason why we should allow indirect imports?

tests/unit/adaptors/graph/test_layers.py
```diff
@@ -958,3 +958,79 @@ def _pairwise(iterable):
     a, b = itertools.tee(iterable)
     next(b, None)
     return zip(a, b)
+
+
+class TestClosedLayers:
```
seddonym, line 963: I think we're missing tests for layers with multiple closed layers, right?

docs/usage.rst
```diff
@@ -409,6 +410,18 @@ Higher level analysis
             ),
         )
 
+    Closed layers
```
seddonym, line 413: Great docs! 👏🏻

rust/src/graph/higher_order_queries.rs
```diff
@@ -208,3 +208,80 @@ impl Graph {
...
+        let mut bottom_layer = FxHashSet::default();
+        bottom_layer.insert(bottom_module);
+
+        let top_level = Level::new(top_layer, false);
+        let middle_level = Level::new(middle_layer, false);
+        let bottom_level = Level::new(bottom_layer, false);
+
+        let levels = vec![top_level, middle_level, bottom_level];
+
+        let permutations = graph.generate_module_permutations(&levels);
```
seddonym, line 255: Coming to this function afresh, makes me think it's not very well named (or at least could do with more documentation) - do you agree?

seddonym (changes requested):
This is great stuff. I have hardly anything to add, other than it would be good to sense check the behaviour when there are multiple closed layers - and add tests for that.

rust/src/graph/higher_order_queries.rs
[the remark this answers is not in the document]
Peter554: > Can you see any reason why we should allow indirect imports?

No - I think allowing indirect imports would be dangerous/wrong. By default I think indirect imports should be disallowed. If we find that for some reason we need that in the future, then we could always add that option in the future (but by default I think indirect imports should be disallowed).

src/grimp/domain/valueobjects.py
Peter554: Done.

src/grimp/domain/valueobjects.py
```diff
@@ -69,12 +69,15 @@ class Layer:
...
-    def __init__(self, *module_tails: str, independent: bool = True) -> None:
+    def __init__(self, *module_tails: str, independent: bool = True, closed: bool = False) -> None:
         # `object.__setattr__` is needed since the dataclass is frozen.
         object.__setattr__(self, "module_tails", set(module_tails))
         object.__setattr__(self, "independent", independent)
+        object.__setattr__(self, "closed", closed)
 
     def __str__(self) -> str:
-        return f"{self.module_tails}, independent={self.independent}"
+        module_tails = sorted(self.module_tails)
```
Peter554, line 82: This `sorted` needed to make the test stable.

tests/unit/adaptors/graph/test_layers.py
Peter554: Maybe. I'm unsure exactly what we want to test here... I've added another unit test case for `generate_module_permutations` for two closed layers - does this cover what you were hoping to test?

Peter554:
Hi @seddonym - this is ready for another look now. Sorry the commit history is a bit messy.

rust/src/graph/higher_order_queries.rs
```diff
@@ -84,7 +84,7 @@ impl Graph {
             )
     }
 
-    fn generate_module_permutations(
+    fn generate_illegal_import_permutations_for_layers(
```
seddonym, line 87: Definitely better - might be worth adding a comment to explain what the tuple corresponds to (e.g. is the first one the source and the second one the module that should not be imported by it)?

rust/src/graph/higher_order_queries.rs
```diff
@@ -327,4 +327,57 @@ mod tests {
             ])
         );
     }
+
+    #[test]
+    fn test_generate_module_permutations_two_closed_layers() {
```
seddonym, line 332: I think we should have Python tests for this too (or instead of) - one of the principles of the project is that the Python tests are pretty comprehensive.

rust/src/graph/higher_order_queries.rs
```diff
@@ -327,4 +327,57 @@ mod tests {
...
+
+        let mut middle_bottom_layer = FxHashSet::default();
+        middle_bottom_layer.insert(middle_bottom_module);
+
+        let mut bottom_layer = FxHashSet::default();
+        bottom_layer.insert(bottom_module);
+
+        let top_level = Level::new(top_layer, false, false);
+        let middle_top_level = Level::new(middle_top_layer, false, true); // Closed
+        let middle_bottom_level = Level::new(middle_bottom_layer, false, true); // Closed
```
seddonym, line 355: Good to have a test for two closed layers together, but feels like we could also do with something along these lines

```
one
two
three (closed)
four
five
six (closed)
seven
eight
```

I'd put that in the Python test though probably.

seddonym (changes requested):
Nearly there - I do think it's important that we have reasonably thorough Python tests for this though. Thanks for all your work on it so far!

rust/src/graph/higher_order_queries.rs
Peter554: Okay sure - python tests extended now.

rust/src/graph/higher_order_queries.rs
Peter554: I tested this case:

```
layers = [
    Layer("highest"),
    Layer("high", closed=True),
    Layer("mid"),
    Layer("low", closed=True),
    Layer("lowest"),
]
```

I'm not sure what benefit we'd get from testing even more layers, so kept it simple (in your example you put two open layers "four" and "five" between the closed layers) 🤔

seddonym (approved):
Looks great. Thanks so much for this!

PR #228, opened 2025-06-19
Implement module expression optional fragments

Peter554:


rust/src/module_expressions.rs
```diff
@@ -37,22 +42,29 @@ impl FromStr for ModuleExpression {
...
+        // Handle optional fragments by expanding into all possible combinations
+        let expanded_expressions = Self::expand_optional_fragments(expression);
 
-        for (part, next_part) in expression.split(".").tuple_windows() {
-            match (part, next_part) {
-                ("*", "**") | ("**", "*") | ("**", "**") => {
-                    return Err(GrimpError::InvalidModuleExpression(expression.to_owned()));
+        let mut patterns = Vec::new();
+        for expanded in &expanded_expressions {
+            // Validate each expanded expression
```
Peter554, line 50: The idea here is that after expanding the module expression, all of the expanded expressions should be valid by the old rules. E.g. `a.b[.**]` expands to `a.b` and `a.b.**`, both of which are valid expressions.

seddonym:
Interesting to see how AI performs, but I'm not sure about the change itself. If someone is doing this in, say, [a forbidden contract](https://import-linter.readthedocs.io/en/stable/contract_types.html#forbidden-modules), can't they just remove `as_packages = False`?

Peter554:
> If someone is doing this in, say, [a forbidden contract](https://import-linter.readthedocs.io/en/stable/contract_types.html#forbidden-modules), can't they just remove `as_packages = False`?

@seddonym Yes they could - but `as_packages = False` is not available everywhere. E.g. in `ignore_imports`. Does that influence your opinion?

Peter554:
> Let's leave this PR open for now - there are a few new contracts in the offing which may make it unnecessary, and plenty of other things that I'd like to concentrate on - in particular speeding up the linter.

Okay sounds good to me - I'll put the PR back to draft for now then and we can revisit this if we see a need for it.

Peter554:
Closing this PR since we aren't sure that we want this and the PR is now rather stale.

PR #232, opened 2025-07-27
Find shortest cycle

K4liber:


seddonym:
We ended up tackling the underlying problem with https://github.com/python-grimp/grimp/pull/258, so I'm going to close this.

PR #238, opened 2025-08-06
Optimize find_shortest_chains

Peter554:
This PR improves the performance of the `find_shortest_chains` function by refactoring the underlying implementation to use a more efficient algorithm. The original implementation used a cartesian product of all possible module pairs, which resulted in an O(n²) complexity and poor performance on large graphs.

## Changes:

1. Extracted common chain-finding logic into a new `_find_shortest_chains` helper function that can be reused across different queries
2. Replaced the duplicated code in `find_illegal_dependencies` with calls to the new helper function
3. Optimized `find_shortest_chains` to use the same algorithm as `find_illegal_dependencies`, resulting in significantly better performance in realistic scenarios.

This optimization is particularly beneficial for large dependency graphs where the O(n²) behavior of the previous implementation could cause performance bottlenecks.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,7 @@ Unreleased
 
 * Add closed layers to layer contract.
 * Rename default repository branch to 'main'.
+* Optimise `find_shortest_chains` query.
 
 3.9 (2025-05-05)
 ----------------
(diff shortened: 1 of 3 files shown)


CHANGELOG.rst
```diff
@@ -7,6 +7,7 @@ Unreleased
 
 * Add closed layers to layer contract.
 * Rename default repository branch to 'main'.
+* Optimise `find_shortest_chains` query.
```
seddonym, line 10: Might be worth mentioning that it might lead to slightly different results in some cases? (I agree with you, the results of `find_shortest_chains` is not formally specified, though I'd like us to get to that point.)

seddonym (commented):
Thanks for looking into this!

It's great to see the improvement for no chains found, but there are several other things which are apparently slower. What are you thoughts? Possibly one for a verbal discussion.

PR #239, opened 2025-08-06
Add more useful benchmark for find_shortest_chains

Peter554:
Preparatory commit cherry-picked from https://github.com/seddonym/grimp/pull/238. This preparatory commit slows the benchmark a lot, but this is okay since the scenario we're benchmarking has now changed, not the actual grimp code.

By putting this onto master first, we will then be able to more clearly see the improvement in https://github.com/seddonym/grimp/pull/238.

### DIFF
--- a/tests/benchmarking/test_benchmarking.py
+++ b/tests/benchmarking/test_benchmarking.py
@@ -464,6 +464,40 @@ class TestFindShortestChains:
         )
         assert result == set()
 
+    def test_chains_found_sparse_imports(self, benchmark):
+        """
+        A test case where the number of import chains is significantly less
+        than the number of module combinations. In this case package `a` has
+        100 modules and package `c` has 100 modules, however the number of import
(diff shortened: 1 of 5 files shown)


seddonym (changes requested):
I think this is a good idea, but I'm concerned that we are losing benchmarking on the `large_graph`, which based on a real codebase. I don't see any harm in having both kinds, so would you mind adding these new benchmarks without removing the other ones?

tests/benchmarking/test_benchmarking.py
```diff
@@ -464,6 +464,40 @@ def test_no_chains(self, large_graph, benchmark):
         )
         assert result == set()
 
+    def test_chains_found_sparse_imports(self, benchmark):
```
seddonym, line 467: Love this comment!

seddonym (approved):

PR #242, opened 2025-08-22
Scanning optimization

seddonym:
This PR introduces a couple of optimizations for scanning very large code bases, based on running samply on a large closed-source code base. On my laptop it takes the time building the graph from around 4s to 3s.

### DIFF
--- a/rust/src/filesystem.rs
+++ b/rust/src/filesystem.rs
@@ -6,6 +6,12 @@ use std::ffi::OsStr;
 use std::fs;
 use std::path::{Path, PathBuf};
 use unindent::unindent;
+use lazy_static::lazy_static;
+
+
+lazy_static! {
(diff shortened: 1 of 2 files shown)


rust/Cargo.toml
```diff
@@ -29,6 +29,7 @@ serde = { version = "1.0", features = ["derive"] }
 serde_yaml = "0.9"
 unindent = "0.2.4"
 encoding_rs = "0.8.35"
+once_cell = "1.21.3"
```
Peter554, line 32: Do we need to add `once_cell`? Can we use `lazy_static` (already a project dependency) or even `std::cell::OnceCell`?

seddonym: Good idea - I've switched to `lazy_static` for consistency.

PR #243, opened 2025-09-05
Upgrade dependencies to latest versions

LilyAcorn:


LilyAcorn:
Hmm, I'm starting to think those benchmarks are just a bit flaky.

PR #249, opened 2025-09-05
Use PyO3's declarative modules

LilyAcorn:
Also refactor `GraphWrapper` and move `scan_for_imports` into the `import_scanning` module.

### DIFF
--- a/rust/src/import_scanning.rs
+++ b/rust/src/import_scanning.rs
@@ -1,5 +1,5 @@
-use crate::errors::GrimpResult;
-use crate::filesystem::FileSystem;
+use crate::errors::{GrimpError, GrimpResult};
+use crate::filesystem::{FileSystem, get_file_system_boxed};
 use crate::module_finding::{FoundPackage, Module};
 use crate::{import_parsing, module_finding};
 use itertools::Itertools;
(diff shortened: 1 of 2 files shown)


rust/src/lib.rs
```diff
@@ -30,21 +30,36 @@ use rustc_hash::FxHashSet;
...
-    m.add("ModuleNotPresent", py.get_type::<ModuleNotPresent>())?;
-    m.add("NoSuchContainer", py.get_type::<NoSuchContainer>())?;
-    m.add(
-        "InvalidModuleExpression",
-        py.get_type::<InvalidModuleExpression>(),
-    )?;
-    m.add("ParseError", py.get_type::<ParseError>())?;
-    m.add("CorruptCache", py.get_type::<CorruptCache>())?;
-    Ok(())
+mod _rustgrimp {
```
seddonym, line 26: Been meaning to do this for a while, thanks!

rust/src/import_scanning.rs
```diff
@@ -350,3 +350,76 @@ pub fn imports_by_module_to_py(
...
+///                                  for analysis.
+/// - include_external_packages:     Whether to include imports of external modules (i.e.
+///                                  modules not contained in modules_by_package_directory)
+///                                  in the results.
+/// - exclude_type_checking_imports: If True, don't include imports behind TYPE_CHECKING guards.
+/// - file_system:                   The file system interface to use. (A BasicFileSystem.)
+///
+/// Returns dict[Module, set[DirectImport]].
+#[pyfunction]
+pub fn scan_for_imports<'py>(
```
seddonym, line 369: Good call. Originally the design was to have `lib.rs` as the Python layer but as you will have noticed, we are no longer sticking to that as the crate has grown.

PR #250, opened 2025-09-05
Refactor graph wrapper

LilyFirefly:
Best reviewed and merged after #249.

### DIFF
--- a/rust/src/graph/mod.rs
+++ b/rust/src/graph/mod.rs
@@ -14,7 +14,7 @@ use std::sync::{LazyLock, RwLock};
 use string_interner::backend::StringBackend;
 use string_interner::{DefaultSymbol, StringInterner};
 
-use crate::errors::{GrimpError, GrimpResult};
+use crate::errors::{GrimpError, GrimpResult, ModuleNotPresent};
 use crate::graph::higher_order_queries::Level;
 use crate::graph::higher_order_queries::PackageDependency as PyPackageDependency;
 use crate::module_expressions::ModuleExpression;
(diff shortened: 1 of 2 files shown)


rust/src/graph/mod.rs
```diff
@@ -649,7 +646,7 @@ struct PyImportDetails {
 }
 
 #[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, new)]
-struct PackageDependency {
+struct PyPackageDependency {
```
seddonym, line 613: Hmm, I've just noticed there is some inconsistency in the way we prefix structs that have a Python / Rust counterpart. In the file system classes, the structs that are exposed to Python have the `Py` prefix (just like the pyo3 classes like PySet, PyDict etc.) But in this module it's the other way around - `PyImportDetails`is the pure-Rust one and `ImportDetails` is the Python one. I think that's confusing...

So I think it would be better to call this one `PackageDependency` and the other one `PyPackageDependency` (and, I guess, swap over `ImportDetails` / `PyImportDetails` too). What do you think?

rust/src/graph/mod.rs
```diff
@@ -201,18 +195,15 @@ impl GraphWrapper {
     }
 }
 
-/// Wrapper around the Graph struct that integrates with Python.
 #[pymethods]
-impl GraphWrapper {
+impl Graph {
```
seddonym, line 195: Not sure about this one. My thinking with having a wrapper class (maybe it should be called `PyGraph`) is so we can potentially free ourselves from the GIL if we need to, at some point, e.g. by passing around immutable references to the graph in pure Rust without having to think about Python.

rust/src/graph/mod.rs
```diff
@@ -87,10 +87,7 @@ impl Graph {
             .iter()
             .map(|name| match self.get_visible_module_by_name(name) {
                 Ok(module) => Ok(module),
-                Err(GrimpError::ModuleNotPresent(_)) => {
-                    Err(GrimpError::NoSuchContainer(name.into()))?
-                }
-                _ => panic!("unexpected error parsing containers"),
```
seddonym: Nice change!

rust/src/graph/mod.rs
LilyAcorn: Fair! I'll remove this bit of the refactor.

PR #253, opened 2025-10-17
Add `nominate_cycle_breakers` method

seddonym:
Adds a `nominate_cycle_breakers` method to the graph.

This will be the backing for Import Linter's acyclic contract type. https://github.com/seddonym/import-linter/issues/133

image

- [ ] Add a summary of changes to `CHANGELOG.rst`.
- [ ] Add your name to `AUTHORS.rst`.
- [ ] Run `just full-check`.

### DIFF
--- a/rust/src/graph/cycle_breakers.rs
+++ b/rust/src/graph/cycle_breakers.rs
@@ -1,0 +1,12 @@
+use crate::errors::GrimpResult;
+use crate::graph::{Graph, ModuleToken};
+use rustc_hash::FxHashSet;
+
+impl Graph {
+    pub fn nominate_cycle_breakers(
+        &self,
+        _package: ModuleToken,
+    ) -> GrimpResult<FxHashSet<(ModuleToken, ModuleToken)>> {
+        Ok(FxHashSet::default())
+    }
+}
--- a/rust/src/graph/mod.rs
+++ b/rust/src/graph/mod.rs
@@ -25,6 +25,7 @@ pub mod hierarchy_queries;
 pub mod higher_order_queries;
 pub mod import_chain_queries;
 
+pub mod cycle_breakers;
 pub(crate) mod pathfinding;
 
 static MODULE_NAMES: LazyLock<RwLock<StringInterner<StringBackend>>> =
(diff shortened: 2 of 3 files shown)


docs/usage.rst
```diff
@@ -505,6 +505,48 @@ Higher level analysis
...
+
+    Multiple sets of cycle breakers can exist for a given package. To arrive at this particular set, the following
+    approach is used:
+
+    1. Create a graph whose nodes are each child of the package.
+    2. For each pair of children, add directed edges corresponding to whether there are imports between those two
+       children (as packages, rather than individual modules). The edges are weighted according to the number of
+       imports they represent.
+    3. Calculate the approximately
+       `minimum weighted feedback arc set <https://en.wikipedia.org/wiki/Feedback_arc_set>`_.
```
K4liber: I can see a risk related to only returning minimum weighted feedback arc set out of all found feedback arc sets. If for example we have a following cycle:
- `root.model` importing from `root.gui` (2 imports -> 2 "points")
- `root.gui` importing from `root.model` (3 imports -> 3 "points")
we are suggesting to remove `root.model` dependencies from `root.gui`, but the other way around makes more sense. Maybe we should return sorted list of feedback arc sets together with "scores". Potentially, we could introduce a parameter "max_feedback_arc_sets".

seddonym: I agree that this approach is will not necessarily find the right imports to remove. But to return all of them could potentially be computationally expensive and the number of potential sets of cycle breakers could be very large. The main aim here is to provide a meaningful, deterministic result in the case of a cyclic graph.

I think there would be scope for further tooling to try to arrive at a sensible layering based on a user's model of the domain. For example they might want to specify a non-exhaustive layering (e.g. 'gui is above models') and then this returns cycle breakers based on that additional constraint. That could be added as an optional parameter in a later version.

By the way, IMO helping users decide on the best layer is would probably belong better in [Impulse](https://impulse-cli.readthedocs.io/en/latest/), rather than Import Linter. That has a visual component and I can imagine providing more of a UI to help find the best layering (and even output a draft Import Linter contract).

PR #259, opened 2025-11-02
Cache: Write data map file from rust

Peter554:
This moves the writing of the data map file into Rust.

Based on https://github.com/python-grimp/grimp/pull/251, with some fixes.

### DIFF
--- a/rust/src/filesystem.rs
+++ b/rust/src/filesystem.rs
@@ -30,13 +30,13 @@ pub trait FileSystem: Send + Sync {
 
 #[derive(Clone)]
 #[pyclass]
-pub struct RealBasicFileSystem {}
+struct RealBasicFileSystem {}
 
 // Implements a BasicFileSystem (defined in grimp.application.ports.filesystem.BasicFileSystem)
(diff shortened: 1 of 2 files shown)


rust/src/caching.rs
```diff
@@ -26,6 +51,54 @@ pub fn read_cache_data_map_file<'py>(
     Ok(imports_by_module_to_py(py, imports_by_module))
 }
 
+#[allow(unused_variables)]
```
seddonym: Any reason for including this attribute?

seddonym (changes requested):
This is great! Thanks for taking the trouble to work on it.

There's just a couple of very minor tidy up things, mostly non-blockers.

seddonym (approved):
Lovely stuff. Thanks so much!

I think concentrating on speeding up cache reads (in particularly, by switching from json to a binary file format) is a particularly worthwhile thing to do, and this work will get us closer to that. Let me know if you end up beginning work on that, just to avoid duplicated effort.

PR #264, opened 2025-11-12
Demo: Build graph rust (Sequential)

Peter554:
Forked from https://github.com/python-grimp/grimp/pull/263. 

In this version we discover all modules before we start parsing any of them. This version is a little less performant, but it is easier to understand and debug.


Peter554:
I'm going to close this one for now - it's stale and we're not actively working on it.

PR #276, opened 2025-12-16
Add deprecation notice for `NamespacePackageEncountered`

### LINKED ISSUE #272: `NamespacePackageEncountered` exception docstring needs revisiting?
This error doesn't seem to be raised anymore in the codebase itself, since namespace packages are now supported at the top level, I think the docstring should probably be updated.

https://github.com/python-grimp/grimp/blob/82ebd22f29f8a73962c5c071c91002cf64879434/src/grimp/exceptions.py#L19-L25

Perhaps something like

```Python
class NamespacePackageEncountered(GrimpException):
    """
    Indicates that there was unexpectedly no __init__.py at the top level.

    This indicates a namespace package (see PEP 420). Previously, Grimp did not support this situation.
    This error is kept for backward-compatibility reasons and is not raised by Grimp itself anymore.
    """
```

nathanjmcdougall:
[x] Add tests for the change. In general, aim for full test coverage at the Python level. Rust tests are optional.
[ ] Add any appropriate documentation.
[x] Add a summary of changes to `CHANGELOG.rst`.
[x] Add your name to `AUTHORS.rst`.
[x] Run `just full-check`.

Resolves #272

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -3,9 +3,10 @@ Changelog
 =========
 
 latest
------------------
+------
 
 * Deprecate the `NamespacePackageEncountered` exception.
+* Improve contribution experience for Windows developers using Just.
 
 3.14 (2025-12-10)
 -----------------
--- a/justfile
+++ b/justfile
@@ -170,3 +170,5 @@ full-check:
     @just build-docs
     @just test-all
     @echo '👍 {{GREEN}} Linting, docs and tests all good.{{NORMAL}}'
+
+set windows-shell := ["powershell.exe", "-NoProfile", "-NoLogo", "-Command"]


src/grimp/exceptions.py
```diff
@@ -17,12 +20,16 @@ class NoSuchContainer(GrimpException):
...
-    """
-    Indicates that there was no __init__.py at the top level.
-
-    This indicates a namespace package (see PEP 420), which is not currently supported. More
-    typically this is just an oversight which can be fixed by adding the __init__.py file.
-    """
+    # https://github.com/python-grimp/grimp/issues/272
+    def __init__(self, *args, **kwargs):
+        warnings.warn(
+            "NamespacePackageEncountered is deprecated; graphs can now be built from "
```
seddonym: What do you think about adding 'This exception will be removed in Grimp 3.16.'?

nathanjmcdougall: I think it's a good idea to give an indication of the deprecation duration, but I think it's better to use a point in time/date, e.g. `This exception will be removed after January 31st 2026`.

This has some advantages (speaking in generalities):

- It's more actionable for the user. I have no immediate visibility or control of when Grimp 3.16 will be released. Maybe next week? Maybe in 6 months?
- As the Grimp maintainer, it gives you more flexibility with your release schedule. If your cadence is higher for whatever reason, then you aren't compelled to try and rush the deprecation.
- By leaving it as a "one-sided" time window (i.e. "removed after") rather than fixing it to the version, it allows you to indefinitely defer it, e.g. if you discover a major stakeholder/group of users needs to more time to migrate, or you're short on time on your personal life as a dev.

Obviously, most of these considerations are pretty moot. It's unlikely to affect many users. I'm just thinking about general principles for deprecation.

nathanjmcdougall: > Though you can restrict your dependencies to be below Grimp 3.16, until you've removed the deprecated exception, right?

That's a good point. I guess probably the ideal is to include both. In this case, I'll keep it simple and stick with just Grimp 3.16

PR #282, opened 2025-12-16
Add the "UP" (pyupgrade) ruff rules

nathanjmcdougall:
[ ] Add tests for the change. In general, aim for full test coverage at the Python level. Rust tests are optional.
[ ] Add any appropriate documentation.
[] Add a summary of changes to `CHANGELOG.rst`.
[ ] Add your name to `AUTHORS.rst`.
[] Run `just full-check`.

As versions of Python get dropped (like Python 3.9 recently), it's good to keep on top of opportunities for syntax changes/modernization. This Ruff rule is an easy to way to do this.

My ulterior motive is to get more Ruff rulesets enabled in the codebase, gradually. To some extent, I'm just testing the waters here about your appetite for this :)

Is there a need for a changelog entry here? I feel it's a bit excessive.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,6 +2,11 @@
 Changelog
 =========
 
+latest
+----------------
+
+* Improve contribution experience for Windows developers using Just.
+
 3.14 (2025-12-10)
 -----------------
 
--- a/justfile
+++ b/justfile
@@ -170,3 +170,5 @@ full-check:
     @just build-docs
     @just test-all
     @echo '👍 {{GREEN}} Linting, docs and tests all good.{{NORMAL}}'
+
+set windows-shell := ["powershell.exe", "-NoProfile", "-NoLogo", "-Command"]


pyproject.toml
```diff
@@ -84,3 +84,5 @@ namespaces = false
 [tool.ruff]
 line-length = 99
 exclude = ['tests/assets'] # These are sample packages for tests to run under - we don't want ruff to mess with them.
+[tool.ruff.lint]
+select = ["E4", "E7", "E9", "F", "UP"]
```
nathanjmcdougall: These other rules are the defaults

https://docs.astral.sh/ruff/configuration/#configuring-ruff

seddonym (approved):
Love it.

> My ulterior motive is to get more Ruff rulesets enabled in the codebase, gradually. To some extent, I'm just testing the waters here about your appetite for this :)

100% on board!

> Is there a need for a changelog entry here? I feel it's a bit excessive.

I agree, it's only important for user-facing changes IMO.

PR #284, opened 2025-12-17
Enable various Ruff rulesets

nathanjmcdougall:
Another testing the waters sort of thing, I'm wanting to make sure:

- this is the right size PR in terms of reviewability (e.g. too many rules in one PR)?
- you're happy with the tweaks & rules themselves

I've included rulesets which are relatively uncontroversial (thus mostly already complied with) and have autofixes enabled. I've ignored specific rules which are too controversial.

To access autofixes with `ruff check . --fix`, I wonder whether it would be helpful to configure Just to run Ruff with `--fix`, if not by default then at least with opt-in for specific use cases, e.g. for within pre-commit.


tests/unit/application/graph/test_chains.py
```diff
@@ -164,7 +165,7 @@ def test_demonstrate_nondeterminism_of_equal_chains(self):
 
         one_chain = (source, a, b, c, destination)
         other_chain = (source, d, e, f, destination)
-        assert (result == one_chain) or (result == other_chain)
+        assert result in (one_chain, other_chain)
```
nathanjmcdougall: Stylistic (similar comment to before regarding unpackaging syntax)

tests/unit/application/test_scanning.py
```diff
@@ -730,13 +730,6 @@ def my_function():
         ("import namespace.foo.green.alpha.one", "namespace.foo.green"),
         ("from namespace.foo.green.alpha import one", "namespace.foo.green"),
         ("from ..green.alpha import one", "namespace.foo.green"),
-        ("from .. import green", "namespace.foo.green"),
```
nathanjmcdougall: These parametrizations are duplicated.

PR #289, opened 2025-12-18
Fix/address flaky tests

### LINKED ISSUE #288: Address flaky tests
I'm finding that `tests\functional\test_error_handling.py::test_syntax_error_includes_module` will fail when run on its own, but not when running the whole test suite.

On other projects I've had good success with [`detect-test-pollution`](https://pypi.org/project/detect-test-pollution/) which bisects the test suite to find leaky tests.

Basically you just run `uv run detect-test-pollution --fuzz --tests ./tests` and then follow the prompts.

Typical workflow looks like this:

<details>

```
> uv run --with detect-test-pollution detect-test-pollution --fuzz --tests ./tests
discovering all tests...
-> discovered 852 tests!
run 1...
-> found failing test!
try `detect-test-pollution --failing-test tests/functional/test_build_and_use_graph.py::TestFindDownstreamModules::test_as_package_true --tests ./tests`!
>
> uv run --with detect-test-pollution detect-test-pollution --failing-test tests/functional/test_build_and_use_graph.py::TestFindDownstreamModules::test_as_package_true --tests ./tests
discovering all tests...
-> discovered 852 tests!
ensuring test passes by itself...
-> OK!
ensuring test fails with test group...
-> OK!
running step 1:
- 851 tests remaining (about 10 steps)
running step 2:
- 425 tests remaining (about 9 steps)
running step 3:
- 212 tests remaining (about 8 steps)
running step 4:
- 106 tests remaining (about 7 steps)
running step 5:
- 53 tests remaining (about 6 steps)
running step 6:
- 27 tests remaining (about 5 steps)
running step 7:
- 13 tests remaining (about 4 steps)
running step 8:
- 7 tests remaining (about 3 steps)
running step 9:
- 3 tests remaining (about 2 steps)
double checking we found it...
-> the polluting test is: tests/unit/adaptors/test_caching.py::TestCacheFileNamer::test_names_meta_cache_per_package
```

</details>

I plan to open a PR to:

- Add `detect-test-pollution` as a dev dependency
- Add `just fuzz-test-pollution-python` as a justfile entry
- Fix the flaky tests so that all subsets of the test suite pass

nathanjmcdougall:
Resolves #288

### DIFF
--- a/tests/unit/test_exceptions.py
+++ b/tests/unit/test_exceptions.py
@@ -60,5 +60,5 @@ class TestSourceSyntaxError:
 
 class TestNamespacePackageEncountered:
     def test_deprecated_access(self):
-        with pytest.warns(DeprecationWarning):
+        with pytest.warns(DeprecationWarning, match="NamespacePackageEncountered is deprecated"):
             exceptions.NamespacePackageEncountered


tests/functional/test_error_handling.py
```diff
@@ -1,14 +1,15 @@
...
 import pytest  # type: ignore
 
 from grimp import build_graph, exceptions
 
 
 def test_syntax_error_includes_module():
-    dirname = os.path.dirname(__file__)
-    filename = os.path.abspath(
-        os.path.join(dirname, "..", "assets", "syntaxerrorpackage", "foo", "one.py")
+    filename = str(
```
seddonym: Out of interest, why does this change fix the issue? They look equivalent to me.

nathanjmcdougall: There's are differences in behaviour between `os.path.abspath` and `pathlib.resolve()`. Specifically (in this case) the latter will normalize Windows drive letter names to capitalize them, e.g. `c:\` -> `C:\`. There are differences with unix too: `abspath` will only make the path absolute, e.g. it won't apply resolutions (e.g. symlinks etc.).

It looks like `build_graph` already applies this normalization, so to get the test to match, we need to match the exact string including capitalization.

Without the fix:

```Python
__file__ == 'c:\\Users\\namc\\Repositories\\grimp\\tests\\functional\\test_error_handling.py'
filename == 'C:\\Users\\namc\\Repositories\\grimp\\tests\\assets\\syntaxerrorpackage\\foo\\one.py'
```

Versus with the fix

```Python
__file__ == 'c:\\Users\\namc\\Repositories\\grimp\\tests\\functional\\test_error_handling.py'
filename == 'c:\\Users\\namc\\Repositories\\grimp\\tests\\assets\\syntaxerrorpackage\\foo\\one.py'
```

When running pytest from the CLI/Just, the VS Code console uses capital `C:\` for the current directory (so tests pass) but when running from the VS Code test-running UI, by default the `--root-dir` argument is set using a non-normalized path with lower-case `c:\`, and the test was failing.

```
['-p', 'vscode_pytest', '--rootdir=c:\\Users\\namc\\Repositories\\grimp', '--capture=no', 'c:\\Users\\namc\\Repositories\\grimp\\tests\\functional\\test_error_handling.py::test_syntax_error_includes_module']
```

PR #290, opened 2025-12-19
Select and comply with all Ruff rules, minus a curated ignore list

nathanjmcdougall:
Continuing from #282 and #284

### DIFF
--- a/tests/functional/test_error_handling.py
+++ b/tests/functional/test_error_handling.py
@@ -1,4 +1,4 @@
-import os
+from pathlib import Path
 
 import pytest
 
@@ -6,9 +6,10 @@ from grimp import build_graph, exceptions
 
 
 def test_syntax_error_includes_module():
-    dirname = os.path.dirname(__file__)
-    filename = os.path.abspath(
-        os.path.join(dirname, "..", "assets", "syntaxerrorpackage", "foo", "one.py")
+    filename = str(
+        (
+            Path(__file__).parent.parent / "assets" / "syntaxerrorpackage" / "foo" / "one.py"
+        ).resolve()
     )
 
     with pytest.raises(exceptions.SourceSyntaxError) as excinfo:
--- a/tests/unit/conftest.py
+++ b/tests/unit/conftest.py
@@ -1,14 +1,15 @@
 import pytest
 
 from grimp.adaptors.modulefinder import ModuleFinder
-from grimp.application.config import settings
 from grimp.application.graph import ImportGraph
+from tests.config import override_settings
 
 
 @pytest.fixture(scope="module", autouse=True)
 def configure_unit_tests():
-    settings.configure(
+    with override_settings(
         IMPORT_GRAPH_CLASS=ImportGraph,
         MODULE_FINDER=ModuleFinder(),
         FILE_SYSTEM=None,
-    )
+    ):
+        yield


src/grimp/adaptors/caching.py
```diff
@@ -1,15 +1,19 @@
-import hashlib
+from __future__ import annotations
```
nathanjmcdougall: This helps avoid quoted annotations

src/grimp/adaptors/caching.py
```diff
@@ -93,7 +97,8 @@ def setup(
         )
         cache._build_mtime_map()
         cache._build_data_map()
-        assert cache.cache_dir
+        if not cache.cache_dir:
```
nathanjmcdougall: This is from rule S101 which is perhaps a bit opinionated

https://docs.astral.sh/ruff/rules/assert/

src/grimp/adaptors/caching.py
```diff
@@ -104,7 +109,7 @@ def read_imports(self, module_file: ModuleFile) -> set[DirectImport]:
         try:
             cached_mtime = self._mtime_map[module_file.module.name]
         except KeyError:
-            raise CacheMiss
+            raise CacheMiss from None
```
nathanjmcdougall: https://docs.astral.sh/ruff/rules/raise-without-from-inside-except/

src/grimp/adaptors/modulefinder.py
```diff
@@ -1,10 +1,17 @@
...
-from collections.abc import Iterable, Set
+from typing import TYPE_CHECKING
 
 from grimp.application.ports import modulefinder
-from grimp.application.ports.filesystem import AbstractFileSystem
 from grimp.domain.valueobjects import Module
 
+if TYPE_CHECKING:
+    from collections.abc import Iterable
+    from collections.abc import Set as AbstractSet
```
nathanjmcdougall: https://docs.astral.sh/ruff/rules/unaliased-collections-abc-set-import/

src/grimp/application/config.py
```diff
@@ -1,19 +1,21 @@
...
-    def configure(self, **config_dict: Any):
+    def configure(self, **config_dict: Any) -> None:
         self._config.update(config_dict)
 
-    def __getattr__(self, name):
+    def __getattr__(self, name: str) -> Any:
         if name[:2] != "__":
             return self._config[name]
-        return super().__getattr__(name)
+        return super().__getattribute__(name)
```
nathanjmcdougall: `Settings` has no parent class, so it inherits from `object`, which defines `__getattribute__`, not `__getattr__`.

pyproject.toml
```diff
@@ -84,6 +84,11 @@ exclude = [
...
-lint.ignore = ["C408", "PLR2004", "PLR0913", "PT006", "PT007", "SIM108", "SIM118", "SIM300"]
-lint.per-file-ignores."tests/**/*.py" = [ "RUF012", "RUF059" ]
+lint.select = [ "ALL" ]
+lint.ignore = [ "A005", "ANN401" ,"ARG002", "C408", "COM812", "D1", "D2", "E501", "FBT", "FIX", "G004", "N818", "PERF203", "PLR2004", "PLR0913", "PLW1641", "PTH", "PT006", "PT007", "RET504", "SIM108", "SIM118", "SIM300", "TD", "TID252" ]
+lint.per-file-ignores."tests/**/*.py" = [ "ANN", "ARG", "B007", "B018", "B033", "D", "EM", "FBT", "N806", "RET505", "RUF012", "RUF059", "S", "TRY" ]
+lint.pydocstyle.convention = "google"
+lint.future-annotations = true
+lint.flake8-builtins.strict-checking = true
+lint.flake8-type-checking.quote-annotations = true
+lint.typing-extensions = false
```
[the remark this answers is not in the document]
seddonym: Happy to consider one, I don't think it's a big deal to add back in, especially if it's only as a dev dependency. typing-extensions gets installed anyway as a dev dependency, via other deps.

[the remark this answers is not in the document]
nathanjmcdougall: Yeah I don't think it's worth it

pyproject.toml
```diff
@@ -84,6 +84,11 @@ exclude = [
     'tests/assets', # These are sample packages for tests to run under - we don't want ruff to mess with them.
     'docs/', # Documentation configuration scripts
 ]
-lint.select = ["A", "C4", "E4", "E7", "E9", "F", "FLY", "FURB", "INP", "PLE", "PLR", "PT", "RUF", "SIM", "UP"]
-lint.ignore = ["C408", "PLR2004", "PLR0913", "PT006", "PT007", "SIM108", "SIM118", "SIM300"]
-lint.per-file-ignores."tests/**/*.py" = [ "RUF012", "RUF059" ]
+lint.select = [ "ALL" ]
```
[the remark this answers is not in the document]
nathanjmcdougall: Cool, good thought re: `pre-commit` config.

src/grimp/adaptors/caching.py
```diff
@@ -1,15 +1,19 @@
...
-from grimp.application.ports.filesystem import BasicFileSystem
-from grimp.application.ports.modulefinder import FoundPackage, ModuleFile
-from grimp.domain.valueobjects import DirectImport, Module
+from grimp import _rustgrimp as rust  # type: ignore[attr-defined]
 
 from ..application.ports.caching import Cache as AbstractCache
 from ..application.ports.caching import CacheMiss
-from grimp import _rustgrimp as rust  # type: ignore[attr-defined]
+
+if TYPE_CHECKING:
```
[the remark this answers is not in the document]
nathanjmcdougall: The TC ruleset from `flake8-type-checking`, specifically codes TC001, TC002, and TC003.

https://docs.astral.sh/ruff/rules/#flake8-type-checking-tc

The runtime penalty is probably near-zero because `import grimp` will basically import all submodules, and grimp has no dependencies. So it would only be stdlib imports but I think it's not worth being concerned about those.

I've gone ahead and reverted this from the PR. We can always add it back in later if there's a desire for it, it's easy to migrate with autofix.

PR #295, opened 2026-01-19
Make package FIPS compatible by marking blake2 hashing with `usedforsecurity=False`

z4y4ts:
- [x] Add a summary of changes to the `latest` section at the top of `CHANGELOG.rst`. (If it's not there, add it.)
- [x] Run `just full-check`.

# Why this change?
In FIPS (Federal Information Processing Standards) compliant environments, Python's hashlib restricts the use of non-FIPS-approved algorithms like BLAKE2 when they might be used for security purposes. However, grimp only uses BLAKE2 to generate safe cache filenames—not for any cryptographic or security purpose.

By explicitly marking the hash with usedforsecurity=False, we tell Python that this hash is not being used for security, allowing grimp to work correctly in FIPS mode without triggering restrictions on non-approved algorithms.

### DIFF
--- a/src/grimp/adaptors/caching.py
+++ b/src/grimp/adaptors/caching.py
@@ -37,7 +37,9 @@ class CacheFileNamer:
         # Use a hash algorithm with a limited size to avoid cache filenames that are too long
         # the filesystem, which can happen if there are more than a few root packages
         # being analyzed.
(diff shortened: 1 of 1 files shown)


seddonym:
Thanks for this - today I learned about `usedforsecurity`.

Issue #297, opened 2026-01-22
Make it compatible with Free Threading (aka no-gil)


thepabloaguilar:
Since Python 3.14 the no-gil feature is not experimental anymore but optional, when some of the dependencies are not marked that can run safely in those versions Python interpreter fallback to the default mode with GIL. This can impact when running on production but also when running tests:
```text
<frozen importlib._bootstrap>:491: RuntimeWarning: The global interpreter lock (GIL) has been enabled to load module 'grimp._rustgrimp', which has not declared that it can run safely without the GIL. To override this behavior and keep the GIL disabled (at your own risk), run with PYTHON_GIL=0 or -Xgil=0.
```

It would be nice if this lib could be compatible with no-gil, if needed I can try to help!

seddonym:
Thanks for the issue, I would like to support free threading but am not sure what is involved or how much work it is.

Any input on that welcome!

thepabloaguilar:
We may want to add a test like this later:

```python
import sys
import sysconfig

# Just importing should be enough for the test
import grimp

# it should only run for Python 3.14t+
def test_ensure_disabled_gil():
    assert sysconfig.get_config_var("Py_GIL_DISABLED") == 1:
    assert sys._is_gil_enabled() is False
```

Source: [Identifying free-threaded Python](https://docs.python.org/3/howto/free-threading-python.html#identifying-free-threaded-python)

thepabloaguilar:
Other interesting resources:
- [Supporting Free-Threaded CPython](https://pyo3.rs/v0.23.4/free-threading.html#supporting-free-threaded-cpython)
- [Python Free-Threading Guide](https://py-free-threading.github.io)

PR #299, opened 2026-01-24
chore: enable experimental free gil test

thepabloaguilar:
This PRs adds Python version "3.14 __t__" to the test matrix, that version is the free GIL version, I've added it as experimental so if something does not work now or in the future from incoming changes it won't break the pipeline.

This is a good start to try making this compliant with free GIL version because:
- We can see where we need to work on
- Where breaking compatibility changes were added

But of course it doesn't mean if nothing is raised during test this lib is already compliant, it's just a first step towards compatibility.

- [ ] Add a summary of changes to the `latest` section at the top of `CHANGELOG.rst`. (If it's not there, add it.)
- [ ] Run `just full-check`.

Ref #297

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -39,9 +39,14 @@ jobs:
     strategy:
       matrix:
         python-version: [
-          "3.10", "3.11", "3.12", "3.13", "3.14"
+          "3.10", "3.11", "3.12", "3.13", "3.14", "3.14t"
         ]
         os: [ubuntu-latest, macos-latest, windows-latest]
+        include:
+          - python-version: "3.14t"
+            experimental: true
+
+    continue-on-error: ${{ matrix.experimental == true }}
 
     steps:
       - uses: actions/checkout@v4
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -60,6 +60,8 @@ jobs:
           python-version: ${{ matrix.python-version }}
 
       - name: Test Python
+        env:
+          PYTHON_GIL: ${{ matrix.experimental == true && '0' || '1' }}
         run: just test-python
 
       - name: Test Rust
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -12,3 +12,4 @@ Authors
 * Nicholas Bunn - https://github.com/NicholasBunn
 * Nathan McDougall - https://github.com/nathanjmcdougall
 * Oleksandr Zaiats - https://github.com/z4y4ts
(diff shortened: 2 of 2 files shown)


thepabloaguilar:
We're going need to work on the rust part:
```text
<frozen importlib._bootstrap>:491: RuntimeWarning: The global interpreter lock (GIL) has been enabled to load module 'grimp._rustgrimp', which has not declared that it can run safely without the GIL. To override this behavior and keep the GIL disabled (at your own risk), run with PYTHON_GIL=0 or -Xgil=0.
```

thepabloaguilar:
Maybe we can set the `PYTHON_GIL` conditionally to `0` only for `3.14t` and see what happens

thepabloaguilar:
It seems even forcing free GIL the tests continue passing, that's may be a great indicator to start

seddonym:
Thanks for this, seems sensible.

I would expect that in its current form, the extension module isn't compatible with freethreading as it interacts with Python objects, e.g. in [this function](https://github.com/python-grimp/grimp/blob/main/rust/src/import_scanning.rs#L388). I would expect it to be quite a lot of work to change this.

Given that under the hood Grimp uses multithreading at the Rust level, I wonder if there is any way we could just apply some kind of global lock, for the time being, so Grimp can't be used across threads, but at least it doesn't say it's incompatible. This is new territory for me so feel free to educate me on what other libraries are doing! 😄

thepabloaguilar:
Hello @seddonym, the _PyO3_ has a guide on how to be compatible with Free GIL: [Supporting Free-Threaded CPython](https://pyo3.rs/v0.23.4/free-threading.html#supporting-free-threaded-cpython)

I haven't went deeper in that guide yet but I think there's no problem we having multithread stuff inside Rust the problem is when modifying objects that are shared with Python runtime without any locks-like mechanisms, this other part of the guide is worth reading: [[pyclass] thread safety](https://pyo3.rs/v0.23.4/class/thread-safety#pyclass-thread-safety). It also covers `#[pymethods]` which also might be related to the mechanisms for `#[pyfunction]`

seddonym:
> PyO3 has a guide on how to be compatible with Free GIL: [Supporting Free-Threaded CPython](https://pyo3.rs/v0.23.4/free-threading.html#supporting-free-threaded-cpython)

Yeah, that was what I read which made me think it would be a lot of work! 😄

seddonym:
It seems to work now out of the box, https://github.com/python-grimp/grimp/pull/310. I'm going to close this one.

PR #303, opened 2026-06-29
Raise ModuleNotPresent for absent modules in direct-import queries (#113)

### LINKED ISSUE #113: Searching on 'yaml' generates a KeyError
To produce the error:

```python
import grimp
graph = grimp.build_graph("nifty_library", include_external_packages=True)
graph.find_modules_that_directly_import('yaml')
```

This results in this error:

```
----> 1 graph.find_modules_that_directly_import('yaml')

File ~/.virtualenvs/kraken-core/lib/python3.10/site-packages/grimp/adaptors/graph.py:234, in ImportGraph.find_modules_that_directly_import(self, module)
    233 def find_modules_that_directly_import(self, module: str) -> Set[str]:
--> 234     return self._importers_by_imported[module]

KeyError: 'yaml'
```

HarperZ9:
Fixes #113.

## Problem

`ImportGraph.find_modules_that_directly_import` handled an absent module
inconsistently with the rest of the API: it silently returned an empty `set()`
(masking the missing module), while the underlying Rust call panics for an
absent module, as noted by the existing `# TODO`.

Its sibling `find_modules_directly_imported_by` had no guard at all and would
surface the raw Rust panic for an absent module.

Other query/manipulation methods (`squash_module`, `is_module_squashed`, ...)
already raise `grimp.exceptions.ModuleNotPresent` in this situation.

## Change

Apply the existing `ModuleNotPresent` guard pattern to both
`find_modules_that_directly_import` and `find_modules_directly_imported_by`, so
an absent module raises `ModuleNotPresent` with the same message format used
elsewhere, and add docstrings documenting the behaviour.

## Behaviour change

`find_modules_that_directly_import` previously returned `set()` for an absent
module; it now raises `ModuleNotPresent`, matching the issue title and the rest
of the API. No internal callers relied on the previous empty-set behaviour.

## Tests

- Added two regression tests in `tests/unit/application/graph/test_direct_imports.py`.
- `pytest tests/unit tests/functional` -> 828 passed, 1 skipped.
- `ruff check`, `ruff format --check`, and `mypy` on the changed files all pass.
- Added a `CHANGELOG.rst` entry under `latest`.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,6 +10,9 @@ latest
 * Tweak Just commands for running version-specific Python tests.
 * Remove `typing-extensions` as a dependency.
 * Make package FIPS compatible by marking blake2 hashing with `usedforsecurity=False`.
+* Raise `ModuleNotPresent` instead of an unclear error when `find_modules_that_directly_import` or
+  `find_modules_directly_imported_by` is called with a module that is not in the graph.
+  (https://github.com/python-grimp/grimp/issues/113)
 
 3.14 (2025-12-10)
 -----------------
--- a/src/grimp/application/graph.py
+++ b/src/grimp/application/graph.py
@@ -217,13 +217,26 @@ class ImportGraph:
         )
 
     def find_modules_directly_imported_by(self, module: str) -> set[str]:
+        """
+        Find all modules that are directly imported by the supplied module.
+
+        If the module is not present in the graph, grimp.exceptions.ModuleNotPresent will be raised.
+        """
+        if not self._rustgraph.contains_module(module):
+            msg = f'"{module}" not present in the graph.'
+            raise ModuleNotPresent(msg)
         return self._rustgraph.find_modules_directly_imported_by(module)
 
     def find_modules_that_directly_import(self, module: str) -> set[str]:
-        if self._rustgraph.contains_module(module):
-            # TODO panics if module isn't in modules.
-            return self._rustgraph.find_modules_that_directly_import(module)
-        return set()
+        """
+        Find all modules that directly import the supplied module.
+
+        If the module is not present in the graph, grimp.exceptions.ModuleNotPresent will be raised.
+        """
+        if not self._rustgraph.contains_module(module):
+            msg = f'"{module}" not present in the graph.'
+            raise ModuleNotPresent(msg)
+        return self._rustgraph.find_modules_that_directly_import(module)
 
     def get_import_details(self, *, importer: str, imported: str) -> list[DetailedImport]:
         """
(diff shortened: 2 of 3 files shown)


HarperZ9:
Heads up on the two red checks, both look unrelated to this change.

The lint failure is `FURB110` on `tests/adaptors/filesystem.py:51` and `:52` (ternary to `or`). This PR does not touch that file, so it looks like a pre-existing full-repo lint hit, probably a ruff version bump enabling `FURB110`. I'm happy to include the two-line `--fix` as a separate commit here if you'd prefer a green run, or leave it out to keep this PR focused.

The `Run tests / Python 3.11 macOS` failure is in the Rust build (`linking with cc failed`, `_PyDict_GetItemWithError` from pyo3), a native-link/runner issue rather than anything in Python. The other matrix cells pass.

The change itself is Python only (the `ModuleNotPresent` guard plus tests and a changelog entry), and `pytest tests/unit tests/functional` is green locally (828 passed, 1 skipped). Let me know if you'd like me to fold in the lint fix.

HarperZ9:
Heads-up on the red CI here: it is not caused by this change. `main` currently fails the `Lint and check docs build` job on two pre-existing lint issues from tool version bumps (ruff `FURB110` in `tests/adaptors/filesystem.py` and clippy `collapsible_match` in `rust/src/import_parsing.rs`), and because that job shares the workflow with the test matrix under `fail-fast`, it cancels the test jobs too. #305 fixes both. Once that lands I will rebase this PR so its checks go green. The change itself is unrelated to the lint failures.

HarperZ9:
Correction to my note above: I said "the other matrix cells pass," which the now-completed run contradicts, so let me give the accurate picture.

The only genuine failure is the **macOS 3.11** job, which fails in the native build step (`maturin build_editable` → `linking with cc failed`, `_PyDict_GetItemWithError`). With fail-fast, that canceled the rest of the matrix, so the other jobs show "fail" but were canceled mid-run rather than failing (e.g. the ubuntu 3.11 job got through `test_error_handling.py` and then logged `The operation was canceled`).

Why this isn't from this PR: the change is pure Python (the `ModuleNotPresent` guard + tests + a changelog line), `main`'s latest CI run is green, and a pure-Python change can't affect the Rust/pyo3 link. It looks like a transient/toolchain pyo3 issue on macOS (there's also a failed Dependabot pyo3 bump on `main`). Locally `pytest tests/unit tests/functional` is green (828 passed, 1 skipped).

The one thing that would actually clear it is a re-run of the failed jobs, which I can't trigger from a fork — would you mind kicking one off? Separately, the lint job trips `FURB110` in `tests/adaptors/filesystem.py` (untouched by this PR, looks like a newer ruff enabling the rule repo-wide); happy to include the two-line `--fix` here or leave it for a separate cleanup, your call.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

seddonym:
Thanks - I'll get to this in the next few days.

seddonym:
Thanks so much!

PR #304, opened 2026-06-29
Raise ValueError when a module is assigned to more than one layer (#165)

### LINKED ISSUE #165: PanicError when duplicate layers are passed to find_illegal_dependencies
E.g.

```
>>> import grimp
>>> g = grimp.build_graph("importlinter")
>>> graph.find_illegal_dependencies_for_layers(layers=("grimp.domain", "grimp.domain"))
thread '<unnamed>' panicked at src/importgraph.rs:144:63:
no entry found for key
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/.../grimp/adaptors/graph.py", line 372, in find_illegal_dependencies_for_layers
    return _layers.find_illegal_dependencies(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/.../grimp/adaptors/_layers.py", line 46, in find_illegal_dependencies
    rust_package_dependency_tuple = rust.find_illegal_dependencies(
                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
pyo3_runtime.PanicException: no entry found for key
```

HarperZ9:
Fixes #165.

## Problem

`find_illegal_dependencies_for_layers` terminates with an opaque Rust panic when the same module is assigned to more than one layer:

```
>>> graph.find_illegal_dependencies_for_layers(layers=("grimp.domain", "grimp.domain"))
thread '<unnamed>' panicked at src/importgraph.rs:144:63:
no entry found for key
pyo3_runtime.PanicException: no entry found for key
```

A module cannot sit at two different levels at once, so this is an input error that should be reported clearly rather than surfacing as a panic from the extension.

## Change

Add a Python-side guard (`_check_layers_do_not_share_modules`) that runs after `_parse_layers` and raises a descriptive `ValueError` naming the repeated module(s) before the Rust call:

```
ValueError: Modules can only belong to one layer, but the following appear in more than one: 'grimp.domain'.
```

It works on the parsed `Layer.module_tails`, so it covers the simple-string, `set`, and `Layer` input forms.

## Tests

- Added `TestInvalidLayers` in `tests/unit/application/graph/test_layers.py` for the string and `Layer` duplicate cases.
- `pytest tests/unit tests/functional` -> 828 passed, 1 skipped.
- `ruff check`, `ruff format --check`, and `mypy` on the changed files pass.
- Added a `CHANGELOG.rst` entry under `latest`.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,6 +13,8 @@ latest
 * Raise `ModuleNotPresent` instead of an unclear error when `find_modules_that_directly_import` or
   `find_modules_directly_imported_by` is called with a module that is not in the graph.
   (https://github.com/python-grimp/grimp/issues/113)
+* Raise `ValueError` instead of pyo3_runtime.PanicException when a duplicate layer
+  module is passed to `find_illegal_dependencies_for_layers`.
 
 3.14 (2025-12-10)
 -----------------
--- a/src/grimp/application/graph.py
+++ b/src/grimp/application/graph.py
@@ -446,9 +446,11 @@ class ImportGraph:
             lower layer (the 'upstream') to the higher layer (the 'downstream').
 
         Raises:
+            ValueError: if the same module is assigned to more than one layer.
             NoSuchContainer: if the container is not a module in the graph.
         """
         layers = _parse_layers(layers)
+        _check_layers_do_not_share_modules(layers)
         try:
             result = self._rustgraph.find_illegal_dependencies_for_layers(
                 layers=tuple(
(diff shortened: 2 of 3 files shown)


HarperZ9:
Heads-up on the red CI here: it is not caused by this change. `main` currently fails the `Lint and check docs build` job on two pre-existing lint issues from tool version bumps (ruff `FURB110` in `tests/adaptors/filesystem.py` and clippy `collapsible_match` in `rust/src/import_parsing.rs`), and because that job shares the workflow with the test matrix under `fail-fast`, it cancels the test jobs too. #305 fixes both. Once that lands I will rebase this PR so its checks go green. The change itself is unrelated to the lint failures.

seddonym:
Thanks for the PR! I'll have a look in the next few days.

seddonym:
Thanks for your contribution, and for the other ones! 👏🏻  All merged now.

PR #305, opened 2026-06-30
Fix the lint failures that are failing CI on main (ruff + clippy)

HarperZ9:
## What

The `Lint and check docs build` job is red on `main`, which (because it shares the workflow with the test matrix and `fail-fast` is on) cancels the test jobs too, so every PR shows all-red even though the Python suite passes. The lint job runs `just lint` = `lint-python` then `lint-rust`, and both halves currently fail from independent tool version bumps:

### 1. Python: ruff `FURB110`

`ruff.toml` uses `lint.select = ["ALL"]`, which enables `FURB110` ("Replace ternary `if` expression with `or` operator"), flagging:

```
FURB110  tests/adaptors/filesystem.py:51:28
FURB110  tests/adaptors/filesystem.py:52:44
```

Fix: replace `x if x else {}` with the idiomatic `x or {}` (the rule's prescribed fix; semantically identical for the falsy-to-`{}` default).

### 2. Rust: clippy `collapsible_match`

With the Python lint fixed, `just lint` proceeds to `lint-rust`, where `cargo clippy --all-targets --all-features -- -D warnings` fails on clippy 1.96's `collapsible_match` in `rust/src/import_parsing.rs:131`: the `Expr::Attribute` match arm wraps an `if expr.attr.id == "TYPE_CHECKING" { ... } else { walk_stmt(...) }`.

Fix: collapse it into a pattern guard (clippy's own suggested fix). The `else` branch was identical to the `_ =>` fallthrough, so when the guard is false the statement is still walked. Behavior is unchanged.

## Verification

Python, with the pinned `ruff==0.14.8`:

```
$ ruff check                # All checks passed!
$ ruff format --check       # 64 files already formatted
```

Rust, with the CI toolchain `rust 1.96.0`:

```
$ cargo fmt --check                                          # clean
$ cargo clippy --all-targets --all-features -- -D warnings   # clean
$ cargo check --release --all-targets                        # clean
```

On this PR the **`Lint and check docs build` job now passes**, which is the blocker that was cancelling the test matrix.

## Note on the remaining red

After the lint job goes green the test matrix runs, and `Run tests for Python 3.10, macos-latest` fails during the maturin/pyo3 extension build with a macOS linker error (undefined `_PyExc_*` symbols, `clang: linker command failed`). That is a native-build/environment issue, not something this lint-only change touches, and I do not have permission to re-run jobs on this repo. A maintainer re-run (or a look at the macOS pyo3 build) would confirm.

---
Disclosure: authored with AI assistance (Claude), reviewed and verified before submitting. Happy to adjust if you would rather handle either lint via config (pinning the tools or scoping the rules) instead.


seddonym:
Thanks for the PR - I'll have a look in the next few days.

HarperZ9:
I pushed one small follow-up after the macOS 3.10 job failed on the first run.

The failure was not from the lint changes themselves: `maturin` was invoking Cargo from the repo root with `--manifest-path rust/Cargo.toml`, so Cargo did not read the existing `rust/.cargo/config.toml` macOS linker flags. I moved that config to repo-root `.cargo/config.toml`, which keeps the same flags but makes them visible to the Python editable-build path too.

Verified locally:
- `cargo test --manifest-path rust/Cargo.toml --no-default-features`
- `uv run pytest --benchmark-skip`
- `cargo fmt --check`
- `cargo clippy --all-targets --all-features -- -D warnings`
- `uv run ruff format --check`, `uv run ruff check`, `uv run mypy src/grimp tests`, `uv run lint-imports`

The latest GitHub CI run is green across lint/docs, benchmarks, and the full Python OS/version matrix, including the previously failing macOS 3.10 job: https://github.com/python-grimp/grimp/actions/runs/28556086085

seddonym:
Thanks for the fix, much appreciated!

PR #314, opened 2026-08-27
Add Windows ARM64 build

ndabas:
- [x] Add any appropriate documentation.
- [x] Add a summary of changes to the `latest` section at the top of `CHANGELOG.rst`. (If it's not there, add it.)
- [ ] Run `just full-check`.

As the title says.

The changes look slightly crazy because:
- setup-python (and official CPython) only have binary releases from 3.11 onward.
- maturin expects `aarch64` as the architecture while setup-python wants to see `arm64`.
- There was an existing issue in the release workflow which caused only free-threaded wheels to be built for Windows x86 and x64. [This is a known issue](https://github.com/pyo3/maturin-action#windows-build-the-regular-and-free-threaded-interpreters-in-separate-jobs). So I have added a separate step to build the free-threaded wheels as per the recommended workaround.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -108,12 +108,15 @@ jobs:
             target: x64
           - runner: windows-latest
             target: x86
+          - runner: windows-11-arm
+            target: aarch64
     steps:
       - uses: actions/checkout@v5
       - uses: actions/setup-python@v6
         with:
+          # CPython has no official Windows ARM64 build before 3.11.
           python-version: |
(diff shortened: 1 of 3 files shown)


seddonym (changes requested):
Thanks for your contribution! (You should definitely be listed in the contributors 😄 .)

I'm all for building the Windows ARM64 wheels, but if we're doing that, shouldn't we also be running tests under that architecture in the same way?

